PR #115, opened 2014-09-03
Same command line for ISCC & Compil32

KngStr:
This patch removed ISPPCC project, added Ispp surpport for iscc and compil32.

### DIFF
--- a/Projects/ISCC.dpr
+++ b/Projects/ISCC.dpr
@@ -110,8 +110,8 @@ var
 begin
   if GetConsoleScreenBufferInfo(StdOutHandle, CSBI) then
   begin
-    if Length(S) > CSBI.dwSize.X then
-      Str := Copy(S, 1, CSBI.dwSize.X)
+    if Length(AnsiString(S)) > CSBI.dwSize.X then
+      Str := Copy(AnsiString(S), 1, CSBI.dwSize.X - 4) + '... '
     else
       Str := Format('%-' + IntToStr(CSBI.dwSize.X) + 's', [S]);
   end
@@ -183,7 +183,7 @@ function CompilerCallbackProc(Code: Integer; var Data: TCompilerCallbackData;
 
     Pt := GetCursorPos;
 
-    if Pt.Y <= ProgressPoint.Y then
+    if Pt.Y < ProgressPoint.Y then
       Exit
     else if ProgressPoint.X < 0 then begin
       ProgressPoint := Pt;


KngStr:
Sorry, I'm late.

1. This progress should always uses single line. 
   But MBCS with ANSI get the error length. 
   So it will use two lines if message is too long.
2. Any why the '... '? Why not '...'?
   What I think is beautiful

martijnlaan:
> This progress should always uses single line. But MBCS with ANSI get the error length. So it will use two lines if message is too long.

Couldn't `WriteToStdHandle` accept Unicode instead? That would avoid the length error right?

KngStr:
Actually It can show MBCS. I don't think use Unicode will solve this issue.
What this issue is how many Pixels those character will take in console.

KngStr:
@martijnlaan 

Add some demo for you to understand what I mean.
That's why I change `<` to `<=`
**Test_first_line.cmd**
```
@echo off

set "iscc=.\isfiles-unicode\ISCC.exe"

::echo "if show this line, then you can see then progress"
call "%iscc%" test2.iss /qp

pause
```
**test2.iss**
```
[Setup]
AppName=My Program
AppVersion=1.5
DefaultDirName={pf}\My Program
DefaultGroupName=My Program
UninstallDisplayIcon={app}\MyProg.exe
Compression=lzma2
SolidCompression=yes

[Files]
Source: c:\windows\fonts\*.*; DestDir: "{app}"; Flags: onlyifdoesntexist uninsneveruninstall
```

**Another bug** Something wrong with quotes.
In "Iscc.dpr" 

`
      end else if IsppMode and GetParam(S, 'D') then begin
        if (Pos(';', S) > 0) or (Pos(' ', S) > 0) then
          S := AddQuotes(S);
`
I used `AddQuotes`, but it's wrong. It can't deal with ";".
So It should be `AnsiQuotedStr(S, '"');`

**test.iss**
```
#pragma error dir
```

**Test_d.cmd**
```
::the real dir is E:\Test\`-=[];',.~!@#$%^&()_+}{

set "dir=E:\Test\`-=[];',.~!@#$%%^&()_+}{"
set "iscc=.\isfiles-unicode\ISCC.exe"

call echo "%%dir%%"
call "%iscc%" /ddir="%%dir%%" test.iss

pause
```

Sorry for my mistake.

martijnlaan:
Ok thanks.

About the first thing: it already says `<=` on master? See https://github.com/jrsoftware/issrc/blob/master/Projects/ISCC.dpr#L186

About the second: how come `AddQuotes` can't deal with ";"? Can't check myself at the moment. And if it can't, shouldn't it be changed so it can?

KngStr:
@martijnlaan 

Thanks！
The second is something like this:

`/dtest=my;you /dtest1=we`
`AddQuotes` wouldn't add quotes for  `test`.
It should be `"test=my;you"`. 

`AddQuotes` is used for something has space.
This case is special. Mybe should add something like `ForceQuotes`?

KngStr:
@martijnlaan 

Sorry! The first thing should be `<`. See: https://github.com/KngStr/issrc/commit/612679f4b19d83934458e5719038457963970e42

martijnlaan:
I've fixed it by replacing the quotes stuff with #1 delimiters like I used for the new /J paramter. Can you double check my changes? TIA.

KngStr:
OK. It works fine. :+1: 

BTW. What about  https://github.com/KngStr/issrc/commit/612679f4b19d83934458e5719038457963970e42

Issue #112, opened 2014-08-05
IDE hangs when minimized and file is updated outside


phantasm:
1. Open some ISS file in Inno Setup GUI
2. Minimize Inno Setup window to taskbar
3. Edit this file with some other text editor, then save it
4. Click on Inno Setup taskbar button
5. You will see "Do you want to reload" dialog
6. Regardless of the choice, Inno Setup window will not emerge from taskbar any further. You will need to close it and restart.

![3fe3f-clip-32kb](https://cloud.githubusercontent.com/assets/369127/3817090/a0a09b88-1cd4-11e4-95dd-895da66be81f.jpg)

maldoinc:
The first reference to this issue mistakenly uses an old branch from another PR. 

I believe PR #164 resolves this issue. Please tell me if it also works for you.

PR #164, opened 2015-05-26
Fix ide hanging when file modified outside of it

maldoinc:
This PR fixes issue #112 

I could almost-reliably reproduce the issue at hand.

The problem would be that the message box is created as a modal window, then the main form is restored thus causing it (the message box) to stay behind the main form and causing it to freeze or simply not respond.

If you would drag away the form to reveal the message box then answer it the application starts responding normally.

This PR fixes this anomaly.

### DIFF
--- a/Projects/CmnFunc.pas
+++ b/Projects/CmnFunc.pas
@@ -268,7 +268,8 @@ begin
   TriggerMessageBoxCallbackFunc(Flags, False);
   try
 {$IFDEF IS_D4}
-    Result := MessageBox(Application.MainFormHandle, Text, Caption, Flags);
+    { On Delphi 4+, simply call Application.MessageBox }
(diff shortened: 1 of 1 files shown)


martijnlaan:
Thanks but I can't merge it like this until I fully understand why you changed it like this and why this fixes the problem, especially since this impacts everything and not just the IDE so I can't be sure it doesn't introduce new problems.

Also I wonder why the commit only changes the Unicode version. Does the non Unicode version not have this problem?

maldoinc:
The previous commit did indeed fix the issue but also broke a few things (namely message boxes from forms other than the main one).

To resolve this I reverted that one and added a check via the `IsIconic` function to detect wether the application is minimized and create a messagebox with a taskbar button.

The problem is mostly reproducible only when the application is minimized via the system minimize button.

martijnlaan:
Thanks!

PR #178, opened 2015-09-10
Minor fix

kant:


DeeHants:
Thank you for the update, but the ISL files must not be UTF-8, instead in their [local codepage](http://www.jrsoftware.org/ishelp/topic_unicode.htm).

You've not made any content changes, but if you do, they should be through the language maintainer (listed at the top of the file) who will then submit them to the project.

(On a related note, please use more meaningful commit messages, probably something like "Converted Spanish.isl to UTF-8")

Thanks

PR #180, opened 2015-10-05
updated TBitmapImage with support of GDI compatible 32-bit RGBA bitmaps

shadow-cs:
There are few limitations: the BMP must be in GDI compatible 32-bit format with pre-multiplied alpha. Some installers that relied on `TBitmapImage` background color replacing feature may break if they used 32-bit RGBX format that is identical to RGBA format and thus will use new painting method (in this case the image may have different background or be rendered completely transparent - conversion to 24-bit will fix that, there is no reason to use less compact format anyway). RGBA images generated by some imaging software may not be supported (GIMP uses bitfields compression method that is not natively supported and does not pre-multiply alpha). Use of third-party software may be needed to generate properly formatted input (ie. Pixelformer).

### DIFF
--- a/Components/BitmapImage.pas
+++ b/Components/BitmapImage.pas
@@ -13,15 +13,33 @@ unit BitmapImage;
 
 interface
 
+{$I ..\Projects\VERSION.INC}
+
 uses
   Windows, Controls, Graphics, Classes;
 
 type
+{$IFDEF IS_D12}
+  TAlphaBitmap = TBitmap;
+{$ELSE}
+  {$DEFINE CUSTOM_BITMAP}
+  TAlphaFormat = (afIgnored, afDefined, afPremultiplied);
+  TAlphaBitmap = class(TBitmap)
+  private
+    FAlphaFormat: TAlphaFormat;
+    procedure PreMultiplyAlpha;
+  public
+    procedure Assign(Source: TPersistent); override;
+    procedure LoadFromStream(Stream: TStream); override;
+    property AlphaFormat: TAlphaFormat read FAlphaFormat write FAlphaFormat;
+  end;
+{$ENDIF}
+
   TBitmapImage = class(TGraphicControl)
   private
     FAutoSize: Boolean;
     FBackColor: TColor;
-    FBitmap: TBitmap;
+    FBitmap: TAlphaBitmap;
     FCenter: Boolean;
     FReplaceColor: TColor;
     FReplaceWithColor: TColor;
@@ -35,6 +53,7 @@ type
     procedure SetReplaceColor(Value: TColor);
     procedure SetReplaceWithColor(Value: TColor);
     procedure SetStretch(Value: Boolean);
+    function GetBitmap: TBitmap;
   protected
     function GetPalette: HPALETTE; override;
     procedure Paint; override;
@@ -51,7 +70,7 @@ type
     property DragMode;
     property Enabled;
     property ParentShowHint;
-    property Bitmap: TBitmap read FBitmap write SetBitmap;
+    property Bitmap: TBitmap read GetBitmap write SetBitmap;
     property PopupMenu;
     property ShowHint;
     property Stretch: Boolean read FStretch write SetStretch default False;
@@ -73,6 +92,23 @@ procedure Register;
 
 implementation
 
+{$IFNDEF IS_D6}
+type
+  TBlendFunction = record
+    BlendOp: BYTE;
+    BlendFlags: BYTE;
+    SourceConstantAlpha: BYTE;
+    AlphaFormat: BYTE;
+  end;
+
+const
+  AC_SRC_OVER = $00;
+  AC_SRC_ALPHA = $01;
+
+function AlphaBlend(DC: HDC; p2, p3, p4, p5: Integer; DC6: HDC; p7, p8, p9,
+  p10: Integer; p11: TBlendFunction): BOOL; stdcall; external 'msimg32.dll' name 'AlphaBlend';
+{$ENDIF}
+
 procedure Register;
 begin
   RegisterComponents('JR', [TBitmapImage]);
@@ -83,7 +119,7 @@ begin
   inherited Create(AOwner);
   ControlStyle := ControlStyle + [csReplicatable];
   FBackColor := clBtnFace;
-  FBitmap := TBitmap.Create;
+  FBitmap := TAlphaBitmap.Create;
   FBitmap.OnChange := BitmapChanged;
   FReplaceColor := clNone;
   FReplaceWithColor := clNone;
@@ -163,6 +199,11 @@ begin
   end;
 end;
 
+function TBitmapImage.GetBitmap: TBitmap;
+begin
+  Result := FBitmap;
+end;
+
 function TBitmapImage.GetPalette: HPALETTE;
 begin
   Result := FBitmap.Palette;
@@ -184,7 +225,8 @@ var
 begin
   with Canvas do begin
     R := ClientRect;
-    Is32bit := FBitmap.PixelFormat = pf32bit;
+    Is32bit := (FBitmap.PixelFormat = pf32bit) and
+      (FBitmap.AlphaFormat in [afDefined, afPremultiplied]);
 
     if Stretch then begin
       W := R.Right;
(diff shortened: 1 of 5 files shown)


martijnlaan:
Thanks!

The C code I was talking about has this bit as well:

```c
GetBitmapBits(dib->hBitmap, bufferSize, buffer);

i = 0;
while (i < bufferSize)
{
  buffer[i + 0] = buffer[i + 0] * buffer[i + 3]/255;
  buffer[i + 1] = buffer[i + 1] * buffer[i + 3]/255;
  buffer[i + 2] = buffer[i + 2] * buffer[i + 3]/255;
  i = i + 4;
}

SetBitmapBits(dib->hBitmap, bufferSize, buffer);
```

Should your code not also have this? I'm not sure what the pros and cons are with regards to the alpha pre-multiply requirement. Do you know?

shadow-cs:
There are no "real" pros or cons, WinAPI just expects it to be
pre-multiplied (which yields better performance during painting - since the
alpha blending is really just sumation of two multiplication + divisions
and pre-multiplication eliminates the need of the source pixel
multiplication + division). Since most imaging tools don't generate
compatible bitmaps and you have to use a tool to convert them (that should
support alpha pre-multiplication) there is no reason why the installer
should do that, this would require the assumption that the alpha is not
pre-multiplied which cannot be determined from the image.
But there is also no reason why not to make this assumption and the
conversion may be done. Question is at which point... IMHO the bitmap
loading should do that if it detects 32-bit bitmap. Delphi 2009+ may use
AlphaFormat := afDefined just before loading the bitmap which would result
in alpha pre-multiplication, older versions would need to re-implement this
behavior (I would suggest creating TBitmap sub-class with this added).
Personally I thought you'd like the first option a little more since more
work is done during pre-processing and not at runtime but I'll gladly
implement pre-multiplication to the image loading algorithm.

martijnlaan:
If I understand you correctly there should be a [Setup] section directive to specify what kind of 32 bit BMP is being used, right? Options `none`, 'premultiplied`, and 'defined'? For `none` it should then use the existing code, for 'premultiplied' the new AlphaBlend code like you have it and for 'definied` the code + multiplication like the C code (not using `afDefined` so it works for the non Unicode version). Default would be `none` to avoid breaking existing stuff.

I've been using Axialis IconWorkshop myself to create these RGBA BMPs and they aren't premultiplied.

shadow-cs:
If you don't want to break anything then yes, I suppose this would the safest way to go. I'd just copy some code from new VCL and create `TAlphaBitmap` that adds `AlphaFormat` property for pre 2009 Delphi.

martijnlaan:
Excellent! Was `TBlendFunction` also introduced by Delphi 2009? If you can add a local definition for it for earlier versions that would be great too.

shadow-cs:
I hoped they added it earlier, I don't have Delphi 2007 but I'll try to obtain in and test it. Otherwise I'll just compare the VCL/RTL sources and let you test it when it is ready.

martijnlaan:
Delphi 7 has it. Delphi 3 does not. It's fine with me to just define it if not Delphi 7 or newer. Please use `{$I VERSION.INC}` and `{$IFNDEF IS_D7}` for that.

shadow-cs:
Do pre Delphi 7 have AlphaBlend function? I think we might need to define that as well.

martijnlaan:
Good point: Delphi 2 & 3 both don't.

shadow-cs:
Done... Tested on 3, 6, 7, XE, XE6, XE7. I added the options you suggested and although I couldn't compile the ANSI version (seems all compilers I have are not supported or I'm missing some changes to the RTL/VCL that are needed I suppose :-( ) I created a test project that verified this solution. The only think that needs to be done is add the documentation, I hope you'll do that?

martijnlaan:
Thanks! I'll check it out as soon as I can. I can do the docs yes if you like.

martijnlaan:
I've checked compilation of the ANSI version. The issues
-There's no TBitmap.PixelFormat (and also no pf32bit)
-There's no TBitmap.ScanLine
Perhaps you could avoid needing these?

BTW here's how my C code does the 32 bit detection:
```c
lpbi = (LPBITMAPINFOHEADER)((char*)lpBitmap+sizeof(BITMAPFILEHEADER));
fAlpha = lpbi->biBitCount == 32;
```

shadow-cs:
Which Delphi version did you use? I know D3 has all that. I know how to determine the info by using the header but I cannot access it since Delphi use private FImage object that stores the DIB section. If you need D2 support and it does not have this I'll need more info...

martijnlaan:
ANSI Inno is compiled with Delphi 2.01 + Delphi 3 + Delphi 7. I suppose could send you a Delphi 2.01 run image for testing? It's so old, that should be ok. It doesn't require any installation and you can use the compile.bat from this repo to compile the ANSI version.

shadow-cs:
I see... Why do the `setup.exe` still rely on Delphi 2, is there a reason why Delphi 3 cannot be used instead? VCL update was pretty significant between those two. I was able to find Delphi 2 on oldversion.com but I don't have VCL sources, do you? Delphi 3 `TBitmap` internal implementation is similar to XE7 but D2 had completely different one and without the sources I'll have to reverse-engineer it which will take more time that I'd rather spent on something else ;-). I was not able to compile the `TBitmapImage` with D2 it was missing `CopyPalette` function. Do you use modified VCL to do the build? If so it would be easier to modify the VCL rather than adding some hacks to access the `TBitmap`'s private fields. I could of course convert the bitmap prior loading it from the stream but this adds more complexity (+ memory requirements) that I hoped to avoid.

shadow-cs:
Well I managed to get it working under D2, I had to add some workarounds for the way how `TBitmap` is implemented internally in this version. Using `TAlphaBitmap` outside of this specific use case in D2 could break the painting. Other Delphi versions are not affected. Shall I push this change?

martijnlaan:
To answer all your questions:
ANSI Setup still relies on D2 because that gives the smallest size.
I do have D2 VCL sources yes. If you want them too please send me your email address.
I don't use a modified VCL.
The change sounds good yes. Thanks once more.

shadow-cs:
Understood, I just find it a little funny these days... What percentage difference will it make compared to the data packed into the setup, but I get the point... Just saying ;-)

If you do not use modified VCL how do you handle the `CopyPalette`? Or is it there in you version? I hope the changes I made are not dependent on the 2.0/2.1 (you said you use 2.1 right?), since this would mean same compiler version could have different VCL interface section which is something I thought was prohibited.

martijnlaan:
CopyPalette is there yes, in Graphics.pas. Version is 2.01, not 2.1.

shadow-cs:
I verified the implementation and pushed the changes, I don't think we could do it "much" better than it is now. The only thing that can be improved is properly assigning and clearing the `FPixelFormat` cache. But I doubt anyone will need that. I'm still unable to compile the `setup.exe` so I only tested this using simple test project that does exactly what `setup.exe` do as well (regarding the bitmap).

martijnlaan:
Cool thanks :) What was the problem with compiling? Did you see the instructions in README.md?

shadow-cs:
Argh... I didn't realize that newer Delphi screwed the `RES` files. Now it's OK. And the alpha bitmap displays just fine ;-)

martijnlaan:
Tried it with the ANSI version and it worked well. The only issue I found is that if you use an alpha bitmap it should never disable the backcolor painting part but it still does namely when the bitmap is big enough. Without alpha it could that since you then wouldn't see the backcolor anyway but with alpha you do.

shadow-cs:
Well that depends... If the parent already painted the background (possibly even with another `TControl`) then the image would blend above that, I think this is desired behavior.

If the parent doesn't clear the background then there could be issues where the parts of an image with `alpha < 255` would be blended over and over resulting in the image becoming more and more opaque each paint cycle. Did you observe this behavior? The parent should always clear the background before calling `TBitmapImage.Paint` no?

martijnlaan:
No I didn't observe that. Maybe I'll just make WizardImageBackColor obsolete just like WizardSmallImageBackColor was made (+then still fix the BackColor+alpha case in TBitmapImage).

shadow-cs:
Will you do it or should I? I'm not exactly sure what you mean specifically. Changing the painting or just the properties of the specific image?

martijnlaan:
@shadow-cs Originally you wrote on the newsgroup:
> use of alpha blending with user defined background
How were you planning to define this background? It seems to be this isn't possible atm but you seem to be happy with just having the alpha blending supported. Am I missing something? Are you planning to use [Code] for this part? Can you give an example of a bitmap and color you're planning to use?

shadow-cs:
> How were you planning to define this background? Am I missing something?

Yes, you are (sort of since you basically answered the question in second paragraph). I changed my windows preferences and use different than white `clWindow` color. So basically this is a compatibility thing if someone is either using different settings or some enhanced theme manager that can do who-knows-what.

> `WizardImageBackColor`
Would adding `clWindow` (or possible other selected system colors) be a problem? But my primary concern was the small image anyway. You can always create 24-bit image or just don't use alpha. (Personally I think adding `clWindow` and `clBtnFace` would be a nice touch since these are the two main colors that the installer itself is using as a background).

Btw. you forgot to delete the `ssBitmapAlphaFormat` in the styler.

martijnlaan:
Ok, thanks. I'm gonna make WizardImageBackColor obsolete and instead make it clWindow, just like for the small image. Also will fix the styler, assumed you didn't know about it :)

martijnlaan:
It should be done now. Can you do some testing to check I didn't screw up something? Also, I've listed you as the contributer of this in whatsnew, let me know if you don't like that.

shadow-cs:
From the top of my head I just have one hint: you don't have to redefine the whole `TBitmap` for the unicode version. `TAlphaBitmap = class(TBitmap)` should do since everything that is needed should be there in the RTL. Unless I'm missing something like scripting.

Not saying you've got it wrong just that this may be safer. Will have a closer look tomorrow.

martijnlaan:
The problem with `TAlphaBitmap = class(TBitmap)` is that the `AlphaFormat` property then resides in a different place for the Unicode version which makes it quite a bit harder to expose it to [Code]. So that's why I changed it to how it's now.

shadow-cs:
I thought this might be why you did it like that. You don't care about `TPixelFormat`, I mean shouldn't it be redefined as well to be accessible from `[code]`?

I reviewed the changes and made a few tests and I think everything is **OK**.

Is Inno the only application that uses `TBitmapImage`? If so, then there is no problem, but in case someone (possibly from newer VCL) is using it, there may be issues with the alpha handling and transparency since `TAlphaBitmap` hides the `AlphaFormat` improperly (to fix this **`inherited`**`AlphaFormat` would need to be used in **`read/write`** of the property).

martijnlaan:
Having `PixelFormat` available to [Code] didn't seem that important to me.

Afaik only IS uses `TBitmapImage` but feel free to make any changes, as long as it doesn't break [Code] in either version :)

shadow-cs:
>Having `PixelFormat` available to [Code] didn't seem that important to me.

I think it ain't as well, information about bit depth isn't really useful and id doesn't affect painting given the way it is currently implemented.

>Afaik only IS uses `TBitmapImage` but feel free to make any changes, as long as it doesn't break [Code] in either version :)

In this case I don't think its all that necessary, I'd rather not pollute the code with more `ifdef`s.

I just wanted to be clear on those things...

martijnlaan:
Merged :+1: Thanks! Hope to see more pull requests from you :)

PR #203, opened 2016-01-28
Full support RightToLeft layout for Arabic

Motaz-Alnuweiri:
I finally implement full `RightToLeft` layout for Arabic language


Motaz-Alnuweiri:
There a problem with `TBitmapImage` it's mirrored when add `WS_EX_LAYOUTRTL`

Issue #201, opened 2016-01-25
[Feature Request] Right To Left Layout for Arabic Language


Motaz-Alnuweiri:
### Hello :)

I would like to request this feature to make Setup windows right to left as in Arabic version of the Windows itself, I did it via Windows APIs but I can't access all windows and I think it's best to add it to Setup by default.

#### Problem
In Arabic translations file I added `RightToLeft = Yes`:
```
[LangOptions]
...
RightToLeft=yes
```
and the Setup windows will look like this:
![Old Right To Left Layout Image](http://i1297.photobucket.com/albums/ag34/MotazNew/Inno%20Setup/1_zpsj2xg5lho.jpg)

#### Solution
I wrote a code to make the windows as it should be in Arabic:
```pascal
[Code]
// Author: Motaz Alnuweiri
// Copyright (C) 2016 for Motaz Alnuweiri.

#ifdef UNICODE
  #define AW "W"
#else
  #define AW "A"
#endif

// Windows APIs to Set ExStyle Layout
function SetWindowLong(HWND: HWND; nIndex: Integer; dwNewLong: longint): HWND;
  external 'SetWindowLong{#AW}@user32.dll stdcall';
function GetWindowLong(HWND: HWND; nIndex: Integer): longint;
  external 'GetWindowLong{#AW}@user32.dll stdcall';

const
  // Windows ExStyle (nIndex)
  // URL: https://msdn.microsoft.com/en-us/library/windows/desktop/ff700543%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396
  GWL_EXSTYLE = -20;
  WS_EX_LAYOUTRTL = $00400000;
  WS_EX_NOINHERITLAYOUT = $00100000;
  WS_EX_RIGHT = $00001000;
  WS_EX_RTLREADING = $00002000;

// On Setup Initialize
procedure InitializeWizard();
var
  ExStyle: Longint;

begin
  // Right To Left Layout for Arabic
  If Not WizardForm.RightToLeft Then
  begin
    ExStyle := GetWindowLong(WizardForm.Handle, GWL_EXSTYLE);
    SetWindowLong(WizardForm.Handle, GWL_EXSTYLE, ExStyle Or WS_EX_LAYOUTRTL 
                  Or WS_EX_NOINHERITLAYOUT xOr WS_EX_RIGHT xOr WS_EX_RTLREADING);
  
[...]

PR #205, opened 2016-02-01
Full support RightToLeft layout for Arabic

motaz-alnuweiri:
I finally implement full RightToLeft layout for Arabic language but I faced two problems:
1. `TBitmapImage` is mirrored when set `WS_EX_LAYOUTRTL` the solution is `SetLayout(hdc, 0)` for the canvas as in this link but the image will disappear.
2. `TCustomFolderTreeView` not effeicted with `WS_EX_LAYOUTRTL` and it's still LTR.


martijnlaan:
So is this all done now or are you expecting me to complete work on this?

Motaz-Alnuweiri:
Finally you replied, OK I think we don't need `WS_EX_LEFTSCROLLBAR` and I expecting you to help me on the problems.

For `TNewNotebook` yes it's a problem but let focus on `TBitmapImage` the problem is because you drawing on Canvas by your self the X will be from right when we use `WS_EX_LAYOUTRTL` so what you suggested ?

martijnlaan:
> Finally you replied, OK I think we don't need WS_EX_LEFTSCROLLBAR and I expecting you to help me on the problems.

> For TNewNotebook yes it's a problem but let focus on TBitmapImage the problem is because you drawing on Canvas by your self the X will be from right when we use WS_EX_LAYOUTRTL so what you suggested ?

Sorry to disappoint you, but I really don't know anything about why the RTL stuff works like it does, you will really have to figure things out on your own. Perhaps the author of the most recent RTL changes can help you, I think that was Zaher Dirkey (@zaher ?) but I'm not sure, Jordan handled that at the time.

martijnlaan:
With regards to TFolderTreeView I found this email exchange with Zaher on why things are as they are:

>> TFolderTreeview isnt modified for Bidi support. Was there some special problem with that or just because you didnt get around to that? Or does  your Arabic Windows maybe do something special with treeviews on its own?

> Treeview, ListView, PageControl not support Right To Left by system (Windows) event on XP (not tested in Vista yet). Some time we use Mirror flags but that make problems espicaly in Delphi.

> BTW TFolderTreeView always the folder must LEft To Right there is not problem, windows installed in our system with English pathes without translation, for that folders names must be LTR for example C:\Program Files\My Program maybe "My Program" is optional for programer to be a
Arabic words but with full path we will get confused when mix Arabic and English words, so we keep it LTR.

To me this seems to say that TFolderTreeView should stay LTR, so you can take it of your problem list.

zaher:
Hi
@Motaz-Alnuweiri  Using Windows Mirroring is bad idea, so I worked on the controls to make it RTL, 
Now you just added a Mirror to it to make ProgressBar and TreeView right to left

Folders should keep it LTR becuase "Progam Files" is English even in Arabic Windows

We should not make ProgressBar rtl, it will make used confused.

Is there any component else not RTL?

zaher:
Is your probelm in TNewNotebook?

martijnlaan:
See the original issue (#201), it's about making the title bar RTL I beleive.

zaher:
I remembered why, Mirroring is not supported in Win95.

But, if you to want to add Mirroing you can add it in method TNewProgressBar.CreateParams instead of calling SetBiDiStyles

zaher:
But I don't agree RTL for TCustomFolderTreeView

martijnlaan:
Do you remember why you used ``WS_EX_LEFTSCROLLBAR``?

zaher:
Not sure, but WS_EX_LEFTSCROLLBAR is exists in win95 not belong to mirror that founded in Win98

zaher:
For title bar, please read the comment in 
http://stackoverflow.com/questions/7437386/how-to-mirror-the-titlebar-of-an-msi-installer

Motaz-Alnuweiri:
I will keep that in my mind `TNewsNotebook` not `TNewsNotebook`.

@zaher Mirroring Windows is not bad idea because in Arabic Windows all windows and controls are RTL even `FolderTreeView` and I made all this changed to make Inno Setup look more native in Arabic Windows.

@martijnlaan my big problem is `TBitmapImage` because you use custom paint, As I mentioned before it's mirrored and to fix that I set `SetLayout(hdc, 0)` but the image will disappear because in mirrored windows the left will become right and the right will become left.

zaher:
Mirroring is bad, started from images, lights, shadows lines for 3d controls, the light should be at Left Top corner, when mirroring it will be at right top, that make the a form ugly for who is sensitive for this things like me.

Also it is buggy in another platforms, we do not know what will happen in new versions of windows, we can't trust it. 

Showing English files in FolderTreeView as Right To Left is so bad for users, mixing Arabic and English at same Edit or Tree make it so confused, i totally disagree to make it RTL.

About progress bar, it is no problem if you like to add RTL to it by using Mirroring, but in my opinion Progress bar should follow files names that appear in the same form.

I had over 5000 customers that using my rtl controls.

Motaz-Alnuweiri:
@zaher RLT mirroring is feature in Arabic Windows, I think if it's _**bad**_ and _**buggy**_ Microsoft will not add or use it in the first place o.O .

I made all this changed based on Arabic Windows not on my own opinion

Motaz-Alnuweiri:
@zaher I used `SetBiDiStyles` because it's simple, If I added it to `TNewProgressBar.CreateParams` then I need to know if `RightToLeft` is set to true.

zaher:
Add new Member to it RightToLeft, like as Delphi/Lazarus we have BidiMode in TControl

zaher:
Check FUseRightToLeft in 
https://github.com/Motaz-Alnuweiri/issrc/blob/master/Components/NewCheckListBox.pas

Motaz-Alnuweiri:
To add `FUseRightToLeft` I need to do a lot of works but with `SetBiDiStyles` I add less code and it's so simple.

I fix the `TCustomFolderTreeView` problem and now it's RTL layout, I need your help @martijnlaan for `TBitmapImage` :disappointed: .

Motaz-Alnuweiri:
Thank you @martijnlaan for nothing, I finally fix the `TBitmapImage` problem :+1: . I need to add RTL to `TNewNotebook` and `TPanel` maybe call it `TNewPanel.

@zaher added `WS_EX_LEFTSCROLLBAR` becuse in LTR the scrollbar is `WS_EX_RIGHTSCROLLBAR`.

martijnlaan:
Are you going to add the new commits to this PR? It's not showing any new stuff at the moment.

Motaz-Alnuweiri:
Yes I will add new commits.

zaher:
TImage/TBitmapImage are Graph control, it is using the canvas of parent and it is not a real wincontrol, so it is managed by VCL.
Now, Delphi not support mirroring that come with Windows, it is use its BidiMode drawing the controls

Motaz-Alnuweiri:
OK, so what you suggest to do @zaher ?

PR #206, opened 2016-02-01
Fix compiling with Delphi XE

virgoparna:
Changes necessary to compile Inno Setup Unicode version with Delphi XE (and possibly also XE2 and newer).


martijnlaan:
Thanks :+1:

PR #208, opened 2016-02-03
Fixing problems with pending deletes

virgoparna:
Currently there is a problem, that when deleting file before replacing it results in file being in pending deletes state, then installer does not recognize it correctly. Also problem, when file is already in pending delete state before starting installation. 
One case of such problem is, if exe or dll from installation is opened over network. In that case deletefile succeeds but file is not actually deleted and cannot be replaced. That results in MoveFile failure and Abort, Retry, Ignore dialog (even if file has restartreplace flag).


martijnlaan:
Do you have a reproduction available for this? BTW I checked Delphi 2's versus Delphi 7's implementation of FileExists and they're identical.

virgoparna:
Now checking, Delphi XE has different FileExists then Delphi5.  In this patch was tested with Unicode version of Inno Setup that I compiled with Delphi XE (I do not use ansi version of Inno Setup). But my simple test, that I used to check it out was compiled with Delphi 5. And in both cases it seems, that FileExists detects files, that were in pending delete state. And I now retested it with test program compiled with Delphi XE and got same result.
Problem occurs, when file that is being  deleted is open from network (it is on shared folder).
Test program was:
---------------------------------------------------------------------------------
program test;
{$APPTYPE CONSOLE}
uses
  SysUtils, Windows, Classes;
const
  FILE_NAME = 'e:\EKSPORT\PUTTY.EXE';
begin
  if Windows.DeleteFile(FILE_NAME) then
  begin
    writeln('Deletion succeeded');
    if FileExists(FILE_NAME) then
    begin
      writeln('File still exists');
      if Windows.DeleteFile(FILE_NAME) then
        writeln('Second success')
      else
      begin
        writeln('failure');
        writeln(GetLastError);
      end;
    end;
  end
  else
  begin
    writeln('Deletion failed');
    writeln(GetLastError);
  end;
end.
----------------------------------------------------------------------------------
That E:\EKSPORT was shared folder under the name \\%hostname%\eksport
And, when I executed 
\\%hostname%\eksport\putty.exe
then test program output was:
Deletion succeeded
File still exists
failure
5
Also in that case, if you have Explorer window open on that directory, then the file will be visible until it is closed.

Both when compiled with Delphi 5 and when compiled with Delphi XE.
But when  I execute e:\EKSPORT\PUTTY.EXE then output was:
Deletion failed
5

Actual executable used as file to be open is not important. It might as well be notepad.exe

So I added that FileExists check into DeleteFileRedir and recompiled it with Delphi XE and used it to build our installer and tried it out there. And the I discovered, that when file is already in such state (deletion is failed), then  optimized NewFileExists from CmnFunc2 of Inno Setup source detects that file as deleted and then tries to overwrite it and then gets Access Denied. So that was the reason for patching NewFileExists. My patch should not break optimization in NewFileExists, because as far as I understand, this is about case, when FileExists does not detect, that  file actually exists.

virgoparna:
I could not get code to show up correctly formatted. Sorry, for multiple edits.

martijnlaan:
Ok, couple of things.
-I wouldn't object to ``NewFileExists`` calling ``FileExists`` as a double check like you changed it to, but I don't understand how this helps you in for the above example.
-I do object to ``DeleteFileDir`` doing an extra ``(New?)FileExists`` + returning an error like you changed it to. Instead I suggest you do the extra check *after* ``DeleteFileRedir`` returned. This way it's only done when it's really needed.

virgoparna:
* Additional check in NewFileExists helps to catch situation, when file is in pending delete state already before starting installation. Without it in that situation error will occur, when trying to overwrite file. And at that time it will not set it to restartreplace  
* I guess that I could modify Install.pas for that. changing `while not DeleteFileRedir(DisableFsRedir, DestFile) do begin`  to repeat with additional success variable. But could that situation cause problems somewhere else too? Maybe  other places calling DeleteFileRedir  should check that situation also? How does the uninstall delete files?

virgoparna:
Actually, now that I'm thinking, correct solution would be to switch replace on restart (when restartreplace is set and MoveFile fails with Acces Denied or Share Violation). I'll try to make such changes.

virgoparna:
I reverted previous changes and mage change to install.pas to handle error with movefile. Hopefully I got it correctly.

martijnlaan:
See the above commit: I basically added what you suggested but also added retries for it, just like it does when DeleteFile fails. Can you please check the commit (779eacafa9e7baa85424ab38bb639feda68589a9) as good as you can? TIA.

virgoparna:
Reverted my change and merged upstream. Works. Closing.

martijnlaan:
Thanks :+1: Did you also test the 'normal case', where retries are not needed? And the case where the entry does not have ``restartreplace`` set? Would be great if you could test these too.

virgoparna:
In normal case install just completes, files are installed and no restart is required. And if I removed the `restartreplace`, then it asks standard Abort, Retry, Ignore confirmation after MoveFile fails..

martijnlaan:
> In normal case install just completes, files are installed and no restart is required. And if I removed the restartreplace, then it asks standard Abort, Retry, Ignore confirmation after MoveFile fails.. 

Ok, cool, thanks. And in the latter case you see the retries happening right?

virgoparna:
Log has 4 "The existing file appears to be in use (183). Retrying." messages before showing dialog.

martijnlaan:
:+1:

martijnlaan:
Tweaked the log a bit, see above.

virgoparna:
`MoveFile: The existing file appears to be in use (183). Retrying.`
So it seems to be working.

Issue #224, opened 2016-04-22
Access violation inside ISPP while compiling with space only lines in [Setup]


DeeHants:
Originally reported by Duvernet Vincent in jrsoftware.innosetup

I've stripped their example down to a bare minimum that triggers the crash:
```
[Setup]
 
;The previous line is important and contains a single space character
AppId=ABC123
AppName=MyApp
AppVersion=1.2.3.4
DefaultDirName={pf}\MyApp

[Code]
function InitializeSetup(): Boolean;
begin
  Log('{#emit SetupSetting("AppID")}');
end;
```
```
---------------------------
Compiler Error
---------------------------
Line 11:

[ISPP] Access violation at address 057F017B in module 'ISPP.dll'. Read of address 00000000.
---------------------------
OK   
---------------------------
```

Importantly, the "blank" line after `[Setup]` consists of a single space
character.
Removing this line, commenting it, or moving it down resolves the problem.

DeeHants:
I believe this is down to the check in [`IsppFunc.pas` line 369](https://github.com/jrsoftware/issrc/blob/master/Projects/ISPP/IsppFuncs.pas#L369) that looks for the end of the `[Setup]` section. This check will be performed on any line between `[Setup]` and the desired directive.
A whitespace only line will be passed through `Trim()` returning a 0 character string, and then, without a bounds check, `...[1]` will try and access the first character of the string, triggering an access violation.
The blank line check ([line 366](https://github.com/jrsoftware/issrc/blob/master/Projects/ISPP/IsppFuncs.pas#L366)) does not do a trim so lets whitespace only lines pass through.

martijnlaan:
Merged pull request.

PR #225, opened 2016-04-22
Fixed an access violation with whitespace only [Setup] lines and ISPP's SetupSetting()

DeeHants:
Believed fix for bug #224. I'm not in a position to test this though.


martijnlaan:
Thanks :+1:

PR #256, opened 2018-04-30
Typo fix Install.pas

jogo-:
Successfully fixed some typos ;-)


martijnlaan:
Thanks 👍 It's "succesvol" in Dutch so that's why I keep forgetting the extra s...

PR #259, opened 2018-05-07
Add Unicode ChineseSimplified language support

zenglanmu:
just converted from the original ChineseSimplified.isl file to utf-8, and fixed the LanguageID


martijnlaan:
Thanks but Simplified Chinese has a codepage (936) so the file should not be unicode.
To get the LanguageID changed please see http://jrsoftware.org/files/istrans/: If you have fixes, comments or suggestions about a particular translation, please contact its maintainer (qiming at clault.com).
If the 936 codepage is wrong please also discuss that with the maintainer.

PR #260, opened 2018-05-14
Update Italian language (ANSI)

bovirus:
Updated Italian language (ANSI text format)


martijnlaan:
Thanks but this pull request doesn't contain any actual changes so it's still UTF8. If you can't get it work work with Git then you can also use http://jrsoftware.org/files/istrans/send2.php

bovirus:
I'm quite difficult to understand why.

I just copied/pasted with Notepad++ the ANSI file (I checked that was ANSI).

I will send you with SEND option.

Thanks.

ronaldhoek:
NOTE: when you paste textual changes using the GitHub website, the text is always committed as UTF8!

To commit ANSI text, you need to make a local copy of the repo on your machine (using a Git client) and then commit these changes to GitHub.

bovirus:
@martijnlaan 

Please update the file into the repository using teh file that isent via SEND function

@ronaldhoek

Thanks for the explanation.

martijnlaan:
I already did, see 46f36acd8d0e589f77437c816167e0f7f3ca6ef6
For the full explanation on how to work with Git, see https://github.com/jrsoftware/issrc/blob/master/CONTRIBUTING.md

PR #264, opened 2018-05-26
Update CompSignTools.pas

jogo-:
Augment visible edit width of 'Command of the Sign Tool' InputQuery


martijnlaan:
Sorry but this doesn't work for me, it doesn't find the form so the edit is not adjusted.

Regardless of that I think it would be better to do this in a more regular way? Namely by adding a new form (so a new .dfm), instead of trying to 'hack' into InputQuery like this.

PR #265, opened 2018-06-06
Update isxfunc.xml

jogo-:
Add info about USERPROFILE environment variable


martijnlaan:
Thanks. Can you remove the "On Win32 platform, " bit? I know the function supports non NT systems, but as a whole IS does not so this bit should not be there in the docs. TIA.

jogo-:
Done.

(Topic related to this forum discussion/issue: http://news.jrsoftware.org/read/article.php?id=106168&group=jrsoftware.innosetup#106168)

martijnlaan:
Thanks 👍

PR #267, opened 2018-06-12
Added Italian language strings

bovirus:


bovirus:
Please don't include this update.
There is an error about "è" accented letter. Ignore it.

bovirus:
What's the right standard of text file for .iss file?

I made a copy/paste from Inno Setup Compiler but is seems that there are some problems on accented letters.

bovirus:
[setup.zip](https://github.com/jrsoftware/issrc/files/2094623/setup.zip)

I have problem with accented letters.
Please add you the file.
I attached it to the post.
Please check at the end letetrs like è ò.
Thanks.

martijnlaan:
Thanks but I think this really needs to be handled via Git, maintaining languages is already too much work for me. Did you use the GitHub's web editor to make the changes. That doesn't seem to work indeed: it will auto change things into UTF8. So you'll need to use Git itself and a proper editor like Notepad or Notepad++. See https://github.com/jrsoftware/issrc/blob/master/CONTRIBUTING.md for an explanation about how to setup Git.

Will close the pull requests for now.

PR #268, opened 2018-06-13
Add Italian strings for Inno Setup Preprocessor

bovirus:
Update strings for Inno Setup preprocessor.

I checked again and the accented (in the file that it's UTF-8)  are OK for me.
If you edit the file letters like è/ò are OK.

Is just a compare between original version and new version show accented letters in wrong way.

Pelase apply the pull request.


bovirus:
@martijnlaan 

I checked that you released Inno Setup 5.6.1.

Why didn't you include my string for preprocessor window in the installer?

martijnlaan:
I haven't had time yet to merge & check it but I will do later and won't forget.

PR #269, opened 2018-06-14
Add NewLine preprocessor macro

martinprikryl:
There seems to be no way to emit a new line in the default Pascal-style preprocessor string syntax.
https://stackoverflow.com/q/45999277/850848
This change adds NewLine macro that can be used for emit a new line, to avoid a need to switch to C-style strings just for this.

### DIFF
--- a/Files/ISPPBuiltins.iss
+++ b/Files/ISPPBuiltins.iss
@@ -50,8 +50,13 @@
 //
 #ifndef __POPT_P__
 # define private CStrings
-# pragma parseroption -p+
+#else
+# pragma parseroption -p-
 #endif
+
+#define NewLine "\n"
+#pragma parseroption -p+
+
 //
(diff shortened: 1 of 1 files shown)


martijnlaan:
Thanks!

martinprikryl:
Sorry, to contact you this way, but I didn't find any other way. You are probably aware anyway. But just in case: http://www.jrsoftware.org/ site is down for few days already.

ajtruckle:
I saw too.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

martijnlaan:
Yes thanks I've noticed that too but can't fix it myself and emailed Jordan about it.

martijnlaan:
@martinprikryl The site is working again. It was an issue at the service provider.

PR #273, opened 2018-06-26
Update ispp.xml

jogo-:


martijnlaan:
Thanks. There are many more places in the ISPP help which are written by Alex and where the word ‘the’ is missing. Would you be interested to fix those too?

jogo-:
OK, then I will review the whole ISPP help

martijnlaan:
Very nice, thanks a lot!

PR #274, opened 2018-06-28
Bump SetupID.

dscharrer:
Keeping this version string updated makes it easier for tools like innounp and innoextract to parse the data format.

### DIFF
--- a/Projects/Struct.pas
+++ b/Projects/Struct.pas
@@ -33,7 +33,7 @@ const
     this file it's recommended you change SetupID. Any change will do (like
     changing the letters or numbers), as long as your format is
     unrecognizable by the standard Inno Setup. }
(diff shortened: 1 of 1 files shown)


ajtruckle:
The latest is 5.6.1.

dscharrer:
The setup data format hasn't changed in 5.6.1 though.

martijnlaan:
Nor did it in 5.6.0 right? Or did I forget to bump?

dscharrer:
Commit 342d78bddd5863b2a1c0874f061dc3645b88eae7 changed the number of images stored in the setup headers. This was probably overlooked because it didn't include a change to Struct.pas.

New flags and enum values have also been added to setup headers in other commits but I don't think those change any field sizes.

martijnlaan:
Yes, I overlooked it indeed. Thanks.

PR #285, opened 2018-10-26
Fixes bad argument replacement on some translated messages.

navossoc:
I did just a few corrections on the translations.

Some arguments (%1) had spaces `% 1` or some were inverted `1%`.

Something like this:
image

I did the fix on "binary level", so I'm sure that I have changed only the bytes needed.
I didn't mess with the encoding, whatever they are. So I don't know why github is messing with some other words on the preview, but the final file is right.

**Diff view**
image

**File view**
image

[]'s


martijnlaan:
Thanks1

PR #286, opened 2018-12-05
include search order for ISPP does not match documentation

puetzk:
The behavior I see is that it's <filename> that searches only the current include path, and "filename" that also searches the directory where the current file resides. The observed behavior also matches what I would have expected based on the rest of ISPP being somewhat based on the C preprocessor and it's <system> and "local" include syntax.

The [code that seems to implement the rule](https://github.com/jrsoftware/issrc/blob/389d20804b25c7bc4bead2887e4e848845a901d3/Projects/ISPP/IsppTranslate.pas#L742-L747) is clearly setting IncludePathOnly := True when the <> delimiters *are* used.

So the documentation just appears to be backwards.

### DIFF
--- a/Projects/ISPP/Help/ispp.xml
+++ b/Projects/ISPP/Help/ispp.xml
@@ -242,8 +242,8 @@
 				</syntax>
 				<description>
 					<para>Includes the &translation; of the specified file.</para>
-					<para>If the filename is enclosed in angle brackets, ISPP first searches for the file in the directory where current file resides, then in the directory where the file that included current file resides, and so on. If the file is not found, it is searched on current include path, set via &pragma;, then on the path specified by INCLUDE environment variable.</para>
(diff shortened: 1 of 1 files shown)


puetzk:
See also http://news.jrsoftware.org/read/article.php?id=106852&group=jrsoftware.innosetup#106852

martijnlaan:
@puetzk Fyi, also did a fix to the code for the other issue you reported. Thanks.

puetzk:
Thanks!

Just to link it, looks like that fix is 5386d42e3cd09da0283d3b4f4b5a0e48beca1c8d, only in the https://github.com/jrsoftware/issrc/compare/is-6 branch (which is fine with me)

martijnlaan:
Yes that's the branch of the next version, but thanks for pointing it out anyway.

PR #292, opened 2019-01-16
update lng.file for IS 6.0.0+

AstaRom:


martijnlaan:
Thanks but when the time is right I will send a message to the maintainers.

AstaRom:
The encoding was incorrect, I will resend.

martijnlaan:
Thanks but don't, I will not accept it, see above.

AstaRom:
What are these "maintainers" for, if you can just now update the translation file for the current changes in the project?

AstaRom:
Or do you think that they will translate better than ordinary users who want to contribute to their favorite software?

PR #294, opened 2019-01-20
Update Corsican messages to version 6.0.0+

Patriccollu:
Additionally changed the apostrophe signs and quotation marks on several strings and also removed unnecessary uppercases.


martijnlaan:
Thanks. Will use the file you sent via the website and commit that later.

PR #301, opened 2019-02-12
Improvements of Lithuanian translation. Updated to version 6.0.0+

Tichij:


martijnlaan:
Thanks. For some reason I'm not getting notifications anymore from GitHub so I didn't see this earlier, sorry.

martijnlaan:
I did not merge the Setup\Lithuanian.isl file you added. Firstly because setup.iss doesn't use these anymore (it was an idea which turned out not to work) and secondly because this was only for official translations.

PR #307, opened 2019-05-06
Update README formatting, add links

ferdnyc:
The README wasn't formatting as intended on GitHub, because the numbers in the section titles prevented them from being interpreted as [setext headings](https://github.github.com/gfm/#setext-headings) despite the underlining.

While it would've been possible to retain the numbering by reformatting, e.g.:

```markdown
1 - Getting Started
---------------------
```

I chose to eliminate the numbers, and instead convert the steps of the _first_ section into an ordered list. (Having both the sections and the steps numbered proved to be too busy and confusing.) That section's list item contents were also shifted right one extra space, to account for the insertion of list numbering.

I also added reference links to many of the internal directory and file paths mentioned. For maximum plain-text readability, I used short reference links, and collected the expanded links at the very bottom of the file after an HTML comment. So for example:

```markdown
**Files\islzma.dll**, **Files\islzma32.exe**, **Files\islzma64.exe** - Compiled
by Visual Studio 2005 from the [Projects\Lzma2\Encoder] directory.

<!-- At the end of the file, the link is defined... -->
[Projects\Lzma2\Encoder]: <Projects/Lzma2/Encoder>
```

There's also some other misc cleanup, like correcting the number of projects in the source tree, using a consistent mood (imperative, rather than indicative) for all of the steps in "Getting Started", and adding markup in a couple of places.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ stored in a compiled resource file.
 **Projects\Lzma2\Decoder\ISLzmaDec.obj**, **Projects\Lzma2\Decoder\ISLzma2Dec.obj** -
 See [Projects\Lzma2\Decoder\compiling.txt].
 
-**ISHelp\ISHelpGen\ISHelpGen.exe** - See [ISHelp\ISHelpGen\compile.bat].
+**ISHelp\ISHelpGen\ISHelpGen.exe** - See [ISHelp\compile.bat].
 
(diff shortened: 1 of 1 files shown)


martijnlaan:
Thanks! I undid 6e0db6fa9642eefd88c4484e0bc0b1db9f49f22f during merge. The actual problem was ISHelpGen's compile.bat missing from Git.

ferdnyc:
Hah, that'll do it!

PR #316, opened 2019-08-26
Update Indonesian translation for 6.0.0+

resenzhu:


martijnlaan:
Thanks. I also received an update from the maintainer so I will use that one instead after a question I have about it is answered. Sorry for the slow reply, I was on vacation.

PR #317, opened 2019-09-24
Increase max-length of VersionInfoProductTextVersion  to 100 characters

mlaily:
Hi,

I would like my setup to have the same informational version as my application's main executable.
To do that, I'm using the `VersionInfoProductTextVersion` directive.

The string I want to use looks like this: `2.13.0.3748;5bda2e80a466cd81bed0155f2a18e37c7c8393b0;BuildAgent=MLYDEV`

The problem is that the result is truncated to 50 characters :/

This PR increases the max-length of this property to 100 characters.

Side note: would it be acceptable to set all the properties inside the `StringFileInfo` block to way longer placeholders? (at least 100 characters each, or maybe even 500 or 1000, so that it's not a problem again in the future?)


martijnlaan:
I think 50 should be enough, sorry. Also this pull request is not mergeable since SetupLdrVersion.res has not been updated.

mlaily:
I'm sad you are thinking that.

This sucks, but I will be putting the version in the copyright field then... (it's the only one being sufficiently long...)

Also, I'm not sure what you mean: I did update SetupLdrVersion.res just after creating my PR because I noticed I forgot it.

PR #318, opened 2019-09-24
Update files for UAC bypass and DLL hijacking

lenovoinc:
Added changes to prevent DLL hijacking and UAC bypass issues that were reported by an external researcher against setup packages built for Lenovo support packages.

### DIFF
--- a/Projects/InstFunc.pas
+++ b/Projects/InstFunc.pas
@@ -241,13 +241,65 @@ begin
   Filename := FN;
 end;
 
+function GetSafeTempDir: String;
+var
+  RetDir, WinDir, TempInstDir, TempWinDir : String;
+  ErrorCode: DWORD;
+begin
+  {
+    InnoSetup's IsAdminLoggedOn only checks if User is a member of Admin Group,
+    not if the process has been elevated by UAC
+    (possible to use IsAdminInstallMode instead?)
+
+    Default to the normal TEMP dir to maximize compatibility
+
+
+    Note - To completely fix DLL hijacking, two additional changes may be made to Delphi system PAS files:
+
+    The following locale changes should also be applied to Delphi's System.pas file as well:
+      Comment out this line so that resource locale DLLs are NOT loaded for InnoSetup
+      Result := LoadLanguageList(FileNameBody, GetUILanguages(0));
+      Result := '';
+
+    Optionally, the "delayed" keyword may be added to the following Delphi files:
+      Winapi.Windows.pas, Winapi.CommCtrl.pas, Winapi.Nb30.pas, System.SysUtils.pas
+    Here is a small example sed script:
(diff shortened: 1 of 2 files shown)


martijnlaan:
Please use the forum first to explain each of these changes in detail. Why are they needed and where can it be seen that they conform to MS's guidelines? Please also make sure to reply to Jordan's posts which you haven't done so far.

Also note the commit is not mergeable in this start: the code is asking questions to itself, it is using a non unique name to start with and seemingly unrelated parts of the manifest have been removed.

mfriedrich74:
The forum is an archaic system not usable by a browser, correct? Or has
that changed recently?

shantanugadgil:
@mfriedrich74 I use the "forum" in the NNTP mode, in Thunderbird Email client. So it is quite easy to manage and follow. 😄

mfriedrich74:
Thunder what? ;)
I use chrome, edge, outlook.

mfriedrich74:
Hmm, seems there is no android version in the google play shop.

Any other clients for android?

martijnlaan:
> The forum is an archaic system not usable by a browser, correct?

See http://www.jrsoftware.org/newsgroups.php, the very first link is the browser interface.

mfriedrich74:
No clients mentioned there, i need to research nntp clients. The first link
to the internet 1.0 page is not useful, not readable on mobiles and not
able to reply. Sorry for hijacking.

lenovoinc:
Please note that the code was asking questions to those who may be more knowledgeable about InnoSetup default methods - specifically if there is a better way to determine if the process has been elevated by UAC.  

Nothing unrelated was removed from the manifest, everything is related to DLL hijacking and ensuring that DLLs are only loaded from the \Windows\SYSTEM32 directory.  This includes adding COMCTL32.dll, and removing the Microsoft.Windows.Common-Controls assembly from the manifest.

Thank you for reviewing these changes.

kamura:
Any update about this problem?  I think this is very serious problem and should be fixed immediately.

bogdanpolak:
@martijnlaan Any update on this? Code under this PR looks like easy to read and understand:
1) function `GetSafeTempDir` is verifying do we really have elevated privileges and checking it tying to create new `TempInst` subfolder inside Windows system folder. @lenovoinc: Can you explain why do we rally need to do this?
2) `XPTheme.manifest` looks like very important change against DLL Hijacking.

Martijan: Why do we really need to use mailing list if discussion can be closed here? Can you explain this?

martijnlaan:
The manifest change is part of version 6.0.5 except it's for just relevant dlls. 

Note that this isn't a fix nor is this a vulnerability. See the forum where this has been explained a few times now, with Microsoft themselves explaining that UAC isn't the security feature everyone thinks it is. This also means the other changes achieve nothing. And yes, please use the forum for discussion.

PR #326, opened 2019-10-12
Italian translation update: fixed UTF-8 header and other typos.

albertus82:
@bovirus @martijnlaan From v6.0.3, non-ASCII characters are not displayed correctly when italian language is chosen because the UTF-8 header was accidentally removed from the _isl_ file:

unicode-issue

I also fixed some typos in the translation.

Thanks.

### DIFF
--- a/Files/Languages/Italian.isl
+++ b/Files/Languages/Italian.isl
@@ -1,4 +1,4 @@
-; bovirus@gmail.com
+﻿; bovirus@gmail.com
 ; *** Inno Setup version 6.0.3+ Italian messages ***
 ;
 ; To download user-contributed translations of this file, go to:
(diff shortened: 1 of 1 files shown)


bovirus:
@albertus82 

You just write some minutes before me. :-)

Thanks for the typo fixes.

I believe that the problem about special characters (like accented letters) is not only about Italian language but could be also available on other language that use special characters.

bovirus:
image

image

image

albertus82:
@bovirus You are right. I examined all the _isl_ files and found that also **Corsican.isl** lacks the UTF-8 header. The file was completely transcoded to UTF-8 on the last commit made by @Patriccollu 6 days ago, but the header was not set accordingly. You can easily see what is wrong if you start the Inno Setup installer and choose _Corsu_ language:

unicode-issue

I'm pretty sure we should add the UTF-8 header to this file too.

martijnlaan:
Thanks @albertus82. I will fix Corsican as well.

martijnlaan:
> 
> 
> @albertus82
> 
> You just write some minutes before me. :-)
> 
> Thanks for the typo fixes.
> 
> I believe that the problem about special characters (like accented letters) is not only about Italian language but could be also available on other language that use special characters.

No, special characters work fine. The problem was that Italian.isl was encoded as UTF-8 without a BOM which is a Unix thing and not allowed. Under Windows Unicode files use BOMs. This was now fixed, also for Corsican.

bovirus:
@martijnlaan 

Thanks for the fuix.
Please if you can and you didn't already done to make a new build of the InnoS etup installer with fix for Italian and Corsican language installer strings.

Thanks.

PR #331, opened 2020-02-03
Corrected many punctuation errors in german translation

jensbrand:
I've created a new german translation file with many corrections of punctuation errors (mostly commas). Many thanks to Johnny who made the most work (and who don't want to be named with email)


martijnlaan:
Thanks!

PR #333, opened 2020-02-25
Cleanup.

AstaRom:


martijnlaan:
Thanks but as CONTRIBUTING.md says, mass whitespace changes should not be done.

PR #337, opened 2020-06-19
Add version number helper functions.

uecasm:
Putting my code where [my mouth](https://groups.google.com/forum/#!msg/innosetup/iVueYUTFpSs/AzxL_EwiBAAJ) is.

Properly, the version should be a `UInt64`, but it appears that ROPS doesn't support that at present.  In practice it won't matter as I doubt anyone has ever created something with a major version > 32767.

There's a little bit of a disconnect between the Code and ISPP implementations of `PackVersionComponents` -- the latter is forced to use only 32-bit ints.  The help could be improved a bit by having links between the Code and ISPP topics, but I couldn't find any existing examples of such links so I'm not sure if that's supported or not.

Possible future extension might be to make a alternative version of `GetVersionNumbers` that returns a 64-bit version directly.  Or perhaps introduce a structure representing the four components.

### DIFF
--- a/Files/ISPPBuiltins.iss
+++ b/Files/ISPPBuiltins.iss
@@ -168,6 +168,26 @@
 #define GetFileOriginalFilename(str FileName) GetStringFileInfo(FileName, ORIGINAL_FILENAME)
 #define GetFileProductVersion(str FileName) GetStringFileInfo(FileName, PRODUCT_VERSION)
 //
+// PackVersionComponents
+//
+// Packs individual version components into MS/LS values compatible with GetVersionNumbers.
+//
+#define PackVersionComponents(int Major, int Minor, int Rev, int Build, *MS, *LS) \
+  MS = (Major & 0xFFFF) << 16 | (Minor & 0xFFFF), \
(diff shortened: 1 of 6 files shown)


uecasm:
One possible alternative implementation (that I'm not sure how to write at present, but I'm willing to give it a shot if preferred) would be to make a `TVersion` class that encapsulates all of this conversion and comparison logic.

However comparison might be more annoying with that implementation since as far as I can tell, ROPS doesn't support class operator overloading,  Probably a `function Compare(Other: TVersion): Integer;` method is the best it could manage.

martijnlaan:
Very nice, thanks!

martijnlaan:
> There's a little bit of a disconnect between the Code and ISPP implementations of PackVersionComponents -- the latter is forced to use only 32-bit ints

@uecasm So I took a quick peak at the ISPP source and adding a seperate 64-bit int type looks like quite some work but changing simply changing ISPP's int to ne 64-bit doesn't look too hard. That should be usefull and backward compatible, right?

PR #338, opened 2020-06-19
An incomplete sketch of a version class.

uecasm:
This is an alternate-timeline version of #337, which might be irrelevant now that you've merged that, but I thought I'd share it anyway.

It's incomplete (lacking docs), but it does work.

I'm not entirely happy with it since it requires an explicit heap `Create` in the Inno script (which then also means a `Free` that everyone will forget to write).

But on the plus side, it does allow a bit more natural syntax:

```
  Version := TVersion.Create(1, 2, 3, 4); // or CreateFromNumbers(MyAppMS, MyAppLS)
  // can now access Version.Value, or Version.MS/LS, or Version.Major/Minor/Revision/Build
  // or use Version.ToStr() or Version.Compare(AnotherVersion)
  // assigning to any of the properties updates the other properties as well,
  // so conversions Just Work™
  Version.Free;  // but everyone will forget to do this
```

Unfortunately I didn't see any obvious way to make ROPS accept a non-class with properties.  (But it is a non-class in the native Inno side such that theoretically the installer itself could make use of it anywhere else it wants to deal with version numbers, and get the more efficient version.)

(An overloaded `Create` would also have been nicer; I didn't actually test that but I assume it's not supported.)

### DIFF
--- a/Projects/ScriptClasses_C.pas
+++ b/Projects/ScriptClasses_C.pas
@@ -25,6 +25,25 @@ uses
   uPSC_std, uPSC_classes, uPSC_graphics, uPSC_controls, uPSC_stdctrls,
   uPSC_forms, uPSC_extctrls, uPSC_comobj;
 
+procedure RegisterVersion_C(Cl: TPSPascalCompiler);
+begin
+  with Cl.AddClassN(Cl.FindClass('TObject'), 'TVersion') do
+  begin
+    RegisterProperty('Major', 'Word', iptrw);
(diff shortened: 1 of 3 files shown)


ronaldhoek:
The TVersion record would already be a nice addition.

martijnlaan:
Thanks! I like the procedural version better. I think the barrier to use is lower, also it makes it easier to make [Code] and ISPP consistent. I will look into 64 bit support for ISPP for this.

PR #346, opened 2020-07-21
Create Chinese installer translate files.

ghost:
Create:
- Simplified Chinese translate file
- Traditional Chinese translate file.


martijnlaan:
Thanks but both these translations already exist, as unofficial ones.

PR #347, opened 2020-07-22
Mistakes in #emit documentation

martinprikryl:
I believe that the double {{ }} is a mistake, it should be simply {#MyDestDir}.
Also the parameter name is Source, not Filename


martijnlaan:
Thanks Martin.

PR #367, opened 2020-08-10
Fix typo in IsppDebug.pas

jogo-:


martijnlaan:
I don't why this file is in Git, it isn't used. I will remove it.

PR #380, opened 2020-08-18
Automatically open #include files

martijnlaan:
The Compiler IDE now automatically opens (up to 10) `#include` files in tabs which allow you to edit these files from within the Compiler IDE. The list of `#include` files is updated after opening a new main file and after each compilation. This can be turned off in the options. If the option is not turned off, a new Save All menu item is added to the Files menu.

### DIFF
--- a/Projects/CompForm.pas
+++ b/Projects/CompForm.pas
@@ -274,11 +274,11 @@ type
   private
     { Private declarations }
     FMemos: TList;
-    Memo: TISScintEdit;
+    FMainMemo: TISScintEdit;
     FActiveMemo: TISScintEdit;
-    MemosStyler: TInnoSetupStyler;
+    FMemosStyler: TInnoSetupStyler;
     FCompilerVersion: PCompilerVersionInfo;
-    FFilename: String;
+    FMainFilename: String;
     FFileLastWriteTime: TFileTime; //fixme per memo
     FSaveInUTF8Encoding: Boolean; //fixme per memo
     FMRUFilesMenuItems: array[0..MRUListMaxCount-1] of TMenuItem;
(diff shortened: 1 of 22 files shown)


Projects/CompForm.pas
```diff
@@ -2023,42 +1665,46 @@ procedure TCompileForm.ShowOpenDialog(const Examples: Boolean);
...
 end;
 
-procedure TCompileForm.FSaveEncodingItemClick(Sender: TObject);
+procedure TCompileForm.FSaveAllClick(Sender: TObject);
+var
+  Memo: TCompScintEdit;
 begin
-  FSaveInUTF8Encoding := (Sender = FSaveEncodingUTF8);
+  for Memo in FMemos do
+    if Memo.Used then
```
jordanrussell, line 1696: I believe this should include `and Memo.Modified` to avoid unnecessary timestamp changes on unmodified files, and also because some files may be read-only.

Projects/CompForm.pas
```diff
@@ -2908,53 +2643,51 @@ procedure TCompileForm.MemoUpdateUI(Sender: TObject);
...
         end;
       end;
     end;
-    Memo.SetBraceHighlighting(-1, -1);
+    FActiveMemo.SetBraceHighlighting(-1, -1);
   end;
 
 begin
-  if (FErrorLine < 0) or (Memo.CaretPosition <> FErrorCaretPosition) then
+  if (Sender = FErrorMemo) and (FErrorMemo.ErrorLine < 0) or (FErrorMemo.CaretPosition <> FErrorMemo.ErrorCaretPosition) then
```
jordanrussell, line 2673: Confusing mix of `and` and `or` -- which takes precedence?

martijnlaan: Oops, fixed.

Projects/CompForm.pas
martijnlaan: Ok, changed.

PR #382, opened 2020-10-02
Update isportable.iss

jogo-:
A bit of optimization


martijnlaan:
Yes, that's better, thanks.

PR #385, opened 2020-10-22
Fix RunSectionParameters + UninstallRunSectionParameters

jogo-:


martijnlaan:
Thanks, well spotted.

jogo-:
You're welcome.

I also noticed some section "Flags" were missing around here:
https://github.com/jrsoftware/issrc/blob/8dd6ed3f9798e35725a931b2c004784ed6667b10/Components/ScintStylerInnoSetup.pas#L189

More specifically, the Install/Uninstall Delete section **types**:
```pascal
  DeleteSectionTypes: array[0..2] of TInnoSetupStylerParamInfo = (
    (Name: 'files'),
    (Name: 'filesandordirs'),
    (Name: 'dirifempty'));
```
But I wasn't sure if they were skipped on purpose for a good reason, since it's not strictly standard section Flags.

martijnlaan:
This Type parameter has no autocompletion yet but I can add that easily. If you have more suggestions for things that could use autocompletion please let me know, then I'll check it it's doable. Anything that requries a real parse of the [Code] section isn't btw.

PR #387, opened 2020-10-28
Add latest Windows 10 update version+name

jogo-:
With this release and future releases, the Windows 10 release nomenclature is changing from a year and month pattern (YYMM) to a year and half-year pattern (YYH1, YYH2).
https://docs.microsoft.com/en-us/windows/whats-new/whats-new-windows-10-version-20H2


martijnlaan:
Thanks.

PR #389, opened 2020-11-14
Change Corsican translation for current 6.1.0+

Patriccollu:
Hello Martijn,

I just updated the Corsican translation file to make some changes, especially to add an unbreakable space before all question marks and semicolons to follow the locale ponctuation rules.

On the other hand, I'm not sure if there is (again) a problem with the encoding. When I edit my local file with Notepad++, the encoding is correct (UTF-8 BOM) but when I download the current `Corsican.isl` either from `jrsoftware/issrc/Files/Languages/` repository or the new version from my local `/issrc repository`, the encoding is only UTF-8. I even deleted my local `/issrc` repository and created a brand new branch `issrc_201114` but the problem still occurs.

Please can you check and tell me if you see a problem on your side? If yes, do you have some recommendations for updating?

Thanks,
Patriccollu.

### DIFF
--- a/Files/Languages/Corsican.isl
+++ b/Files/Languages/Corsican.isl
@@ -13,6 +13,7 @@
 ; E-mail: Patrick.Santa-Maria[at]LaPoste.Net
 ;
 ; Changes:
+; November 14th, 2020 - Changes to current version 6.1.0+
 ; July 25th, 2020 - Update to version 6.1.0+
 ; July 1st, 2020 - Update to version 6.0.6+
 ; October 6th, 2019 - Update to version 6.0.3+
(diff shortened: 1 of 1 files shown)


martijnlaan:
Thanks for the update.

There's no BOM in the file you committed. Try downloading https://raw.githubusercontent.com/Patriccollu/issrc/db9d653f4ac13dbf64d160f4aa1fa7524f16f928/Files/Languages/Corsican.isl and you can see this for yourself.

I've added the BOM myself now.

Are you trying to use GitHub's online editor to edit the file?

Patriccollu:
Hello Martijn,

Yes, I use GitHub's online editor to edit the file and I usually replace the previous content by the new one.

I just checked the History of commits for `Corsican.isl` and I can see that we _lost_ the BOM on July 1st, 2020.

Could it be a circumvention to **always keep the first line** of the file and only replace the others? I could also modify the Inno Setup version number but keeping the first characters where the BOM is located.

Thanks,
Patriccollu.

martijnlaan:
You should not use the online editor. If you're not familiar with using GitHub without using the online editor I recommend you use https://jrsoftware.org/files/istrans/send2.php to submit updates instead.

PR #394, opened 2020-12-20
Increase internal VersionStringSize of IsMsiProductInstalled

DomGries:
On Windows 10 the product Visual C++ 2008 SP1 (Security Update – MFC) returns "9.0.30729.6161" proving that the build version is included even though [documentation says only major, minor and patch version is included in MSI ProductVersion](https://docs.microsoft.com/en-us/windows/win32/msi/productversion).

Therefore I would not trust it to only have max of 255 in major and minor either and would recommend to increase the max string size to accomodate "65535.65535.65535.65535" which would be 20 digits + 3 dots + 1 null terminator = 24.

### DIFF
--- a/Projects/Msi.pas
+++ b/Projects/Msi.pas
@@ -65,7 +65,7 @@ begin
     Exit;
   end;
 
-  VersionStringSize := 16;
+  VersionStringSize := 24;
   SetLength(VersionString, VersionStringSize);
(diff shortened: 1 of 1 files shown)


martijnlaan:
Thanks but this is not needed. 16 is just for the first try. If it then returns it needs more a second attempt is made with the then known size. So it supports any size of version number.

DomGries:
Yes you are right but since it is parsed into a Int64 packed version number it cant be longer than 24 anyway so why not allocate the max amount from the start

PR #395, opened 2020-12-26
fix the case of a noun

mukhametzyanov:


martijnlaan:
Thanks. Does this look good @yktoo?

yktoo:
@martijnlaan Yes, agreed. It's not the case but the gender but nevermind :-)

PR #396, opened 2021-01-18
Update isetup.xml

jogo-:
Add clarification to comparing time stamps, the time when the current file (or directory) was last written to.


martijnlaan:
Thanks!

PR #397, opened 2021-02-11
Slovak.isl for 6.1.0+ with improvements and comments by MItems

tSlappy:
Slovak.isl for 6.1.0+ with improvements and comments by MItems


martijnlaan:
Thanks. The file is UTF-8 without a BOM so it doesn't work. I will add a BOM myself but can you also make sure to commit future updates with a BOM?

PR #403, opened 2021-06-04
Update setup.iss

bovirus:
@martijnlaan 

Update setup.iss

Add two Italian translations for [Messages] section
Added italian translation for icon description in Inno Setup program group.

I hope that could be possible to add also add strings for crypt installer window

### DIFF
--- a/setup.iss
+++ b/setup.iss
@@ -78,9 +78,26 @@ Name: english; MessagesFile: "files\Default.isl"
 #expr FindFiles("files\Languages\")
 
 [Messages]
-HelpTextNote=/PORTABLE=1%nEnable portable mode.%n/DOWNLOADISCRYPT=1%nEnable ISCrypt.dll download.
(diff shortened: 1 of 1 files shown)


bovirus:
@martijnlaan 

You didn't should make the setup.iss translation updated. It's a job for translator.
For me you jt's to kake setup .iss as much translatable.

For me the current situation it's a nonsense.

For me you didn't make so big advertising at your product.

You make an installer with your priduct that it's partially multilanguage (decrypty window is only an english).

If you use multilanguage installer you should make your instalelr translatable as much as possible all parts of the installer.
If the translator want to tranlate theya dd relative not engglish string. Otehrwise elave in English.
But you made tehssetup.iss with all instrcution to make it translatable.

Otherwise PLEASE DON?T INCLUDE multilanguage files and leave your instalelr only in ENGLISH.

Because for me like now multilanguage with some part in Eenglish appear as not a professional solution like your sofwtare is.
Just appear as something that could be realized better but currently is not realized fine.

PR #406, opened 2021-08-27
Correcting GetSHA256OfUnicodeString example

martinprikryl:
The hash shown in the example shows a hash of `test`, not `Test`.


martijnlaan:
Thanks! I would prefer the hash to be fixed instead of the string. Could you change this? I’m on vacation atm.

martinprikryl:
@martijnlaan  Done.

martijnlaan:
Thanks!

PR #407, opened 2021-08-30
Make it clear that {ini} DefaultValue can also contain constants

martinprikryl:


martinprikryl:
Tested for https://stackoverflow.com/q/68977034/850848#68979539

martijnlaan:
Thanks. It supports that indeed.

PR #410, opened 2021-10-18
Update Romanian.isl

DukeW:
Changed encoding to UTF8 with BOM.
Fixed special characters.
Fixed some wording/grammatical errors.
Changed CodePage to 0


martijnlaan:
Hi, thanks for this update and sorry for the slow reply.

Youre updated file does not match the current version, could you add the missing strings as well? And also update the first two lines for this? 5.5.3 didn't support UTF8 isl files by the way, only 6.0.0 and above do.

To view the differences between versions 6.0.3+ and 6.1.0+ see 9e03ea4de5b8639937d2c4024ec8582a7e63b048.
To view the differences between versions 6.0.0+ and 6.0.3+ see dfdf02aef168be458b64e77afb20ae53a5b4f2ec.
To view the differences between versions 5.5.3+ and 6.0.0+ see 1d6d6869b225c73cde4d3e0541b87264c36f3954

PR #411, opened 2022-01-24
Add a GitHub workflow

dscho:
This PR adds a GitHub workflow. Unfortunately, there is no version of Delphi available for direct download for use in CI builds. We have two options here:

- use a self-hosted runner with Delphi installed
- copy a minimal subset of Delphi's files onto the runner when building

The latter is less cumbersome to maintain, but due to reasons, we need to add a secret password to the repository and a secret URL to download the minimal issrc build environment. This minimal build environment was created by running the build locally while [the Process Monitor was running](https://docs.microsoft.com/en-us/sysinternals/downloads/procmon), then extracting the list of `.exe`, `.dll` and `.dcu` files that were used according to Process Monitor (not the most robust, but at least it results in a `.zip` file that only weighs ~7.5MB as opposed to naively zipping up all the files into a gigantic 1.2GB `.zip` file).

This GitHub workflow opens the door for asking contributors to test out fixes comfortably, merely by pointing them to build artifacts (which this PR does not yet add, as it is intended to gently ask whether this is a direction the InnoSetup maintainer(s) want to go?).

See https://github.com/dscho/issrc/actions/runs/1739134972 for an example how a successful run would look like.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,50 @@ Inno Setup-specific editing guidelines for the help files
   surround it by `<tt></tt>` so that it's displayed in the Courier New font. This is
   a convention used throughout the help file. Example: `<tt>MinVersion</tt>`
 
+Setting up Continuous Integration
+---------------------------------
+
+Delphi is not offered for download via a public link, and while there is a
+free-of-charge community version, its license terms forbid sharing it with others, or
+even let other developers (outside your direct teammates) use it. However, what _is_
+allowed is to copy the files (or a subset thereof) to another machine for the specific
+purpose of supporting unattended builds.
+
+Inno Setup's source code includes a GitHub workflow that performs such unattended
+builds upon `push` events, it requires some setting up, though.
+
+Note: The following instructions assume that you have a correctly-licensed version
+of Delphi installed into `C:\Program Files (x86(\Embarcadero\Studio\20.0`.
+
+To generate the (encrypted) `.zip` file containing the files needed to build
+Inno Setup, use [7-Zip](https://www.7-zip.org/):
+
+```
+cd C:\Program Files (x86)\Embarcadero\Studio\20.0
+"C:\Program Files\7-Zip\7z.exe" a -mx9 -mem=AES256 -p"<password>" ^
+	%USERPROFILE%\issrc-build-env.zip ^
+	bin\dcc32.exe bin\rlink32.dll bin\lnk*.dll ^
+	lib/win32/release/Sys*.dcu lib/win32/release/*.res ^
+	lib/win32/release/System.*.dcu lib/win32/release/System.Generics.*.dcu ^
+	lib/win32/release/System.Internal.*.dcu lib/win32/release/System.Net.*.dcu ^
+	lib/win32/release/System.Net.HttpClient.*.dcu lib/win32/release/System.Win.*.dcu ^
+	lib/win32/release/Vcl.*.dcu lib/win32/release/Vcl.Imaging.*.dcu ^
+	lib/win32/release/Winapi.*.dcu
+```
+
+Then, upload this somewhere public, e.g. by attaching it to a comment in a
+GitHub issue. After that, add this URL as a new repository
+[secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets)
+(at https://github.com/YOUR-USER-NAME/issrc/settings/secrets/actions), under the name
+`ISSRC_BUILD_ENV_ZIP_URL`, and the password as `ISSRC_BUILD_ENV_ZIP_PASSWORD`.
+
+Finally, indicate that your fork of the repository has those secrets, by adding the
+topic `has-issrc-build-env` (click the gear icon next to the "About" label at
+https://github.com/YOUR-USER-NAME/issrc to add the topic).
+
+Once that's done, you're set! The next time you push a branch to your fork, the
+workflow will be triggered automatically.
+
 <!-- Link references -->
 [CONTRIBUTING.md]: <CONTRIBUTING.md>
 [Projects\Lzma2\Encoder]: <Projects/Lzma2/Encoder>
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,40 @@
+name: build
+
+on:
+  push:
+
+jobs:
+  build:
+    # Only set the topic `has-issrc-build-env` if the secrets are available
+    if: contains(github.event.repository.topics, 'has-issrc-build-env')
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: initialize build environment
+        env:
+          ISSRC_BUILD_ENV_ZIP_PASSWORD: ${{ secrets.ISSRC_BUILD_ENV_ZIP_PASSWORD }}
+          ISSRC_BUILD_ENV_ZIP_URL: ${{ secrets.ISSRC_BUILD_ENV_ZIP_URL }}
+        run: |
+          (New-Object Net.WebClient).DownloadFile($env:ISSRC_BUILD_ENV_ZIP_URL, "issrc-build-env.zip")
+          & "C:\\Program Files\\7-Zip\\7z.exe" x -oissrc-build-env -p"$env:ISSRC_BUILD_ENV_ZIP_PASSWORD" issrc-build-env.zip
+          if (!(Test-Path issrc-build-env\bin\dcc32.exe)) {
+            Write-Host "Failed to extract dcc32.exe"
+            Exit 1
+          }
+          Remove-Item issrc-build-env.zip
+          $DELPHIXEROOT = (Get-Item .\issrc-build-env).FullName
+          "DELPHIXEROOT=$DELPHIXEROOT" |  Out-File -NoNewLine -Encoding ascii -Append "$env:GITHUB_ENV"
+      - name: build ISHelpGen
+        run: |
+          cd ISHelp\ISHelpGen
+          # Using `Out-File` to avoid UTF-16LE-encoded .bat files
+          "set DELPHIXEROOT=$env:DELPHIXEROOT" | Out-File -NoNewline -Encoding ascii compilesettings.bat
+          .\compile.bat
+      - name: build issrc
+        run: |
+          "set DELPHIXEROOT=$env:DELPHIXEROOT" | Out-File -NoNewline -Encoding ascii compilesettings.bat
+          "set HHCEXE=%ProgramFiles(x86)%\HTML Help Workshop\hhc.exe" | Out-File -NoNewline -Encoding ascii ISHelp\compilesettings.bat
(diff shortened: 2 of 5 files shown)


martijnlaan:
Thanks. The Delphi build environment files such as .dcu's aren't redistributable are they? This would mean adding them to this repostistory anyway would be a violation of the Delphi license?

dscho:
> I could copy just a subset, as described e.g. here: https://delphi.fandom.com/wiki/Setting_up_a_Delphi_Build_Machine

I think that https://blogs.embarcadero.com/learn-about-delphi-continuous-integration-with-svn-jenkins-and-dunitx/ explicitly encourages a similar setup for Continuous Integration, too. That Embarcadero themselves advertise this setup suggests that they do _not_ even consider PR builds to constitute giving access to a third-party.

dscho:
And [this post](https://www.ideasawakened.com/post/getting-started-with-ci-cd-using-delphi-and-jenkins-on-windows) by a Delphi MVP suggests the following:

> Note that your Delphi license allows you to copy these files to a separate computer only for the sole use of performing automated builds.

And indeed, the [license agreement](https://www.embarcadero.com/products/rad-studio/rad-studio-eula?aldSet=en-GB) says this:

> 2.4. Command Line Compiler. Licensee may install the command line compiler on a separate computer from the Product itself, provided that the sole purpose of doing so is to allow that computer to perform unattended building of applications. The Command Line Compiler shall not be considered a 'redistributable'.

With that explicitly stated, I feel quite certain that the scheme I suggest is covered by Delphi's license agreement. (I am not a lawyer, of course...)

martijnlaan:
In the case that build artifiacts arent added to this repo I don't understand what this PR is aiming to achieve.

dscho:
Oh, we can totally attach build artifacts to the builds. I didn't do that yet in case you'd reject the entire idea of Continuous Integration.

What this PR currently does is just adding a way to verify that everything builds all right, and for licensing reasons this is done only via `push` but not `pull_request` events.

One might argue that it should also work for `pull_request` events, and that the Delphi license allows it under certain circumstances, but that's not how far this PR goes.

We could easily attach the build artifacts to these runs, if you're okay with that.

martijnlaan:
I don't know what you mean with adding artifacts to a 'build' or adding artifacts to a 'run', but adding them to this repo is definitely a no go and it's not clear to me whether this PR requires that or not.

dscho:
> I don't know what you mean with adding artifacts to a 'build' or adding artifacts to a 'run', but adding them to this repo is definitely a no go and it's not clear to me whether this PR requires that or not.

Let me try to clarify. There are a couple of different sets of files at play here:

- ***The source code***
  This is the set of files that you get when cloning https://github.com/jrsoftware/issrc (it contains more than just the source code, of course, as some pre-generated `.dll` and `.exe` files are part of that repository, too).
- ***The build output***
  This is the set of files that you get when running `./build.bat` (after setting up all the required `compilesettings.bat` files). This includes, but is not limited to, `Output/innosetup-*.exe`. Those are the ones most users will be interested in.
- ***The Delphi compiler***
  This is a set of files required to compile the Pascal source code into the final output. Typically, it is assumed that you have a working Delphi setup, but due to licensing restrictions, it can be a little bit of a hassle to get to one (in contrast to other products, _Delphi's_ Community Edition is hidden behind a "give us your email and we send a sales representative your way" wall).

What I mean by "build artifact" is the second set of files, the "build output", if you want. As far as I can see, there is no problem whatsoever to attach those to the build runs. Since I was not sure whether you are interested in Continuous Integration (and from your responses, I still receive no clarity about that), I did not want to put in the work just yet to figure out _what_ files to offer as build artifacts, [but now I did that work anyway](https://github.com/jrsoftware/issrc/pull/411/commits/c545051e3506c6e5fe421fff7b8fee692c913a29). [Here](https://github.com/dscho/issrc/actions/runs/1745117278) is an example of just that:

image

You can download, say, the `installer.zip` file containing the compiler InnoSetup compiler by clicking on the `installer` artifact. Or the documentation as a `.zip` file by clicking on `ISHelp`.

What _would_ be a problem is if we added the third set of files, the Delphi compiler, to either the repository or to the build artifacts. Which is why I did not do that.

You will notice that the workflow [takes pains to download the Delphi compiler from a secret URL, as an encrypted `.zip` file, that is then unpacked using a secret passphrase](https://github.com/dscho/issrc/actions/runs/1745117278/workflow#L15-L25).

To make use of this workflow, you would therefore not only need to merge this PR, but also generate an encrypted `.zip` file from the same set of files from your own Delphi setup, upload it (e.g. to this PR as an attachment, as I did [over here](https://github.com/dscho/issrc/pull/1#issue-1112374032)), then add that URL as a [repository secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository), add the passphrase required to decrypt the `.zip` file as another repository secret, and then add `has-issrc-build-env` [to your repository topics](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/classifying-your-repository-with-topics).

If you do not want to make use of this workflow, but are okay with me and other contributors to make use of it, all you need is to merge this PR and then forget all about it.

AdamKearn:
> You will notice that the workflow [takes pains to download the Delphi compiler from a secret URL, as an encrypted `.zip` file, that is then unpacked using a secret passphrase](https://github.com/dscho/issrc/actions/runs/1745117278/workflow#L15-L25).

Can you not just get the download link from the website for the compiler...
And then have the GitHub Action download that using something like CURL on initialisation?

dscho:
> Can you not just get the download link from the website for the compiler...

@AdamKearn No, because Embarcadero does not offer the compiler for download via anonymous link. You have to register, personally, and IIRC a sales representative will send you a personalized link.

AdamKearn:
> a sales representative will send you a personalized link.

Ok and does that link ever expire?   Surely that link can be used as an ENV variable in the GitHub Actions.
That can then be used to download the compiler.

dscho:
> > a sales representative will send you a personalized link.
> 
> Ok and does that link ever expire? 

I believe it does expire at some stage.

Also note that there is no link to the compiler alone, only to the full RAD Studio.

Which is at the core of the problem with your suggestion.

> Surely that link can be used as an ENV variable in the GitHub Actions.

If the link was public, or if you were allowed to share it with others, sure. But neither is the case.

> That can then be used to download the compiler.

That would directly violate the license terms, which specifically prohibits licensees from letting a third-party use the software.

As I detailed earlier (with links and quotes, it took a long time to put together, please read it as carefully as I wrote it), in our context only installing for unattended builds is allowed. You still must take all precautions to _not_ violate the terms by making even only that compiler available to others, even inadvertently.

So I still believe that the approach I took is striking the best balance.

martijnlaan:
I'm OK with merging this but could you add a readme explaining how to set it up aimed at those who want to use it? This is now missing. Thanks in advance.

Note the merge would still take some time, I have to check the ISHelpGen commits that are also included, and skip the updated exe and instead do this myself.

dscho:
> I'm OK with merging this but could you add a readme explaining how to set it up aimed at those who want to use it? This is now missing. Thanks in advance.

[Done](https://github.com/jrsoftware/issrc/pull/411/commits/6de6043a389fec8dfecd28a836b525d30726e479).

> Note the merge would still take some time, I have to check the ISHelpGen commits that are also included, and skip the updated exe and instead do this myself.

Oh, there is not actually a _need_ to build and commit that `.exe`, I just included that to prove that the `ISHelpGen` commits lead to a fully-working build product. I dropped that commit.

Here is a [successful run](https://github.com/dscho/issrc/actions/runs/1794239756) of this new GitHub workflow.

martijnlaan:
Thanks. Is it possible to include license.txt in each artifact? And renamed installer to Output? I still see the ISHelpgen changes in this PR btw.

README.md
```diff
@@ -231,6 +231,50 @@ Inno Setup-specific editing guidelines for the help files
...
+Setting up Continuous Integration
+---------------------------------
+
+Delphi is not offered for download via a public link, and while there is a
+free-of-charge community version, its license terms forbid sharing it with others, or
+even let other developers (outside your direct teammates) use it. However, what _is_
+allowed is to copy the files (or a subset thereof) to another machine for the specific
+purpose of supporting unattended builds.
+
+InnoSetup's source code includes a GitHub workflow that performs such unattended
```
martijnlaan, line 243: Should say Inno Setup, so with a space

README.md
```diff
@@ -231,6 +231,50 @@ Inno Setup-specific editing guidelines for the help files
...
+builds upon `push` events, it requires some setting up, though.
+
+Note: The following instructions assume that you have a correctly-licensed version
+of Delphi installed into `C:\Program Files (x86(\Embarcadero\Studio\20.0`.
+
+To generate the (encrypted) `.zip` file containing the files needed to build
+InnoSetup, use [7-Zip](https://www.7-zip.org/):
+
+```
+cd C:\Program Files (x86(\Embarcadero\Studio\20.0
```
martijnlaan, line 253: Typo? Second ( should be a )?

martijnlaan (changes requested):

dscho:
> I still see the ISHelpgen changes in this PR btw.

Yes, the changes are required to build with the current Delphi/MSXML. So I think we really still need those.

What I dropped was the committed `.exe`.

> Is it possible to include license.txt in each artifact?

Yes.

> And rename installer to Output?

Yes.

README.md
martijnlaan: Same thing a bit later btw

martijnlaan:
Why does it have to build ISHelpgen? Can't the workflow use the existing exe? I can see why building ISHelpgen is nice perhaps but it seems unrelated. Could be a separate PR though. One thing that would then have to change it that it would have to be part of build.bat, and not a separate step.

dscho:
> > Is it possible to include license.txt in each artifact?
> 
> Yes.
> 
> > And rename installer to Output?
> 
> Yes.

Done.

> Why does it have to build ISHelpgen?

Because I do not trust pre-commited `.exe` files in general. As a matter of fact, it is probably not the best practice to commit those artifacts in the first place.

> Could be a separate PR though.

Sure. But at some stage I run out of the time I can justify working on this... My main role is maintaining Git for Windows, after all, and my only beef here is that so far, I would not have been able to modify InnoSetup easily in case the Git for Windows installer requires it. So I've spent (actually way more than planned) time on getting a CI going, and am relatively happy with it. If at all, I would say that we should build _even more_ things in the CI build, not less. Things like `MyDll.dll` or `isbunzip.dll`.

But as I said, my time is a bit limited and I think that the current state really is pretty much good enough for now.

README.md
dscho: Fixed.

README.md
dscho: Fixed.

martijnlaan:
Did some fixes related to ISHelpGen and merged it. Let me know if they cause any problems. Thanks!

dscho:
> Did some fixes related to ISHelpGen and merged it. Let me know if they cause any problems. Thanks!

Looks good to me, thanks!

PR #413, opened 2022-03-06
Use UTF-8 with BOM encoding for ChineseTraditional.isl.

iceboy233:


martijnlaan:
Thanks. Please contact the maintainer of the file if you want to make changes. Note that just changing the file to UTF8 doesn't result in actual changes for Setup.

iceboy233:
Note that with UTF-8 (without BOM), the setup will show garbled characters. I think this is https://stackoverflow.com/questions/49922735/inno-setup-unicode-encoding-issue-with-messages-in-iss-script .

PR #417, opened 2022-04-11
Minor fixes

Jens-G:
As the title says. 

The README update only prevents the reader from wasting time following a dead link, it's not a real solution.


martijnlaan:
Thanks.

PR #418, opened 2022-04-11
Max Memos12

Jens-G:
Adds the ability to 
* close single editor files at will (except main and preprocessor)
* lifted the limit of opened include files from 10 to 64 (arbitrary number again)
* improved the TNewTabSet so the user can actually navigate through all of the files

While the solution improves the overall user experience, there's still room left for further improvements.

### DIFF
--- a/Components/NewTabSet.pas
+++ b/Components/NewTabSet.pas
@@ -12,7 +12,7 @@ unit NewTabSet;
 interface
 
 uses
-  Windows, Messages, SysUtils, Classes, Graphics, Controls, Forms, ModernColors;
+  Windows, Messages, SysUtils, Classes, Graphics, Controls, Forms, Menus, ModernColors;
 
 type
   TTabPosition = (tpTop, tpBottom);
@@ -49,6 +49,7 @@ type
     property TabIndex: Integer read FTabIndex write SetTabIndex;
     property Tabs: TStrings read FTabs write SetTabs;
     property TabPosition: TTabPosition read FTabPosition write SetTabPosition default tpBottom;
+    property PopupMenu;
     property OnClick;
   end;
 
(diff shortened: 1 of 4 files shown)


Jens-G:
> [CLA assistant check](https://cla-assistant.io/jrsoftware/issrc?pullRequest=418) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/jrsoftware/issrc?pullRequest=418) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/jrsoftware/issrc?pullRequest=418) it.

Except it does not work. :-(

martijnlaan:
Thanks! Looked at it shortly and looks like a good start but some stuff is missing before this is mergeable. It should at least remember which files are closed (probably by intergrating with the existing KnownIncludedFiles code) and have a way to reopen a file (from the menu?) I think.

Jens-G:
As I wrote: there is still room for improvement.

Jens-G:
> It should at least remember which files are closed

It does, internally.

>  It should probably also reopen automatically on things like errors in a closed memo, or when you debug into code in a closed memo if it doesn't do that already.

Yep, that's a good point. I have to check that.

martijnlaan:
> > It should at least remember which files are closed
> 
> It does, internally.

I meant if you close the IDE, restart it and reopen the same script. It will then automatically show all known include files even before you recompile. This should not show the previously closed files. 

> > It should probably also reopen automatically on things like errors in a closed memo, or when you debug into code in a closed memo if it doesn't do that already.
> 
> Yep, that's a good point. I have to check that.

Thanks.

martijnlaan:
Nice work on the new commits, looks good! Can it also be made to reopen on Step Into debugging? I also wonder it there now should be an option to hide all by default?

One thing though, your coding style (I mean the formatting like whitespace use, TRUE or True, variable naming, etc) is a bit too different from the existing coding style. Would you be willing to bring it closer to the existing coding style? I hope so.

PR #420, opened 2022-08-10
Improve high DPI support for the installer

wangwenx190:
By adding the dpiAwareness field to the manifest file.

As a drive-by, add the recommended fallback value "PerMonitor"
to the dpiAwareness field of the compiler's manifest file.
The dpiAwareness field allows multiple values.

### DIFF
--- a/Projects/Compil32.manifest.txt
+++ b/Projects/Compil32.manifest.txt
@@ -28,7 +28,7 @@
 <application xmlns="urn:schemas-microsoft-com:asm.v3">
     <windowsSettings>
       <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true/pm</dpiAware>
-      <dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">PerMonitorV2</dpiAwareness>
+      <dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">PerMonitorV2, PerMonitor</dpiAwareness>
     </windowsSettings>
 </application>
 <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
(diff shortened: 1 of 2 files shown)


martijnlaan:
Thanks! I think this might not actually be correct for the installer though. Did you try it? The installer implements its own scaling (Scaled = False) and doesnt work like Compil32 which uses VCL's built in scaling (Scaled = True). Also just changing the text files does nothing, the .bat files need to be used to compile the updated text into the .res file. I guess this means you didn't test the changes? Also whatsnew would need an explanation in what way high dpi support is improved.

wangwenx190:
Yes I didn't actually test this PR because I've done this for many applications, and even if the application doesn't respond to the `WM_DPICHANGED` message, it's still a general improvement because it can prevent Windows from stretching your user interface automatically, and this can improve the sharpness of your UI elements.

> The installer implements its own scaling

I don't quite understand how the installer works internally, but from this line this is the perfect case to enable PerMonitor DPI aware. For PerMonitor DPI aware applications, you should implement your own scaling functionalities indeed, for other DPI awareness level , Windows may try to scale your user interface and what's worse, the non-client area will also become blurry. Besides, the original manifest file of the installer also set `dpiAware` to `true`, it means the system DPI awareness level, what I was trying to do is just migrate it to a more modern DPI awareness level, so I don't see what's wrong with this PR. If you really want to turn off Windows's high dpi support, you should set it to `unaware` or just remove the `dpiAware` entry.

> the .bat files need to be used to compile the updated text into the .res file.

OK, if you agree with my above explaination, I'll continue working on this PR.

>  Also whatsnew would need an explanation in what way high dpi support is improved.

OK.

martijnlaan:
The installer is dpi aware so setting it to unaware makes no sense. It however is not aware per monitor so setting it to permonitor is also wrong until monitor awareness is added. The most natural approach to do this is setting Scaled to True and using TImageCollection (both just like Compil32) and then doing a lot of testing and ensuring backward compatibility, especially for pages/controls added by script creators. A PR which sets awareness to per monitor but does not actually add monitor awareness is not acceptable to me.

wangwenx190:
OK, I understand it now. But I'm not familiar with developing delphi applications, so currently I'm not able to add such funtionalities to the installer. So the best choice for now is to close this PR it seems. Really sorry to waste you some time!

PR #422, opened 2022-08-21
Update Korean.isl

VenusGirl:


martijnlaan:
Thanks. Why did you remove the names of all previous contributors? Can you restore that part?

VenusGirl:
As I tried to make it and use it in the installation program, there were so many mistakes such as spelling, wrong translation, etc., and the meaning was not conveyed well, so I translated it completely into the original English without referring to the previous contributor's translation.

Should there be a history of previous contributors?

If that part is necessary and necessary, I will revise it.

VenusGirl (commented):
It is a translation that is as faithful as possible to the original English text without omission.

martijnlaan:
Thanks. I'm currently away from home and will merge this PR when I get back.

kristinWalker1991 (approved):

kristinWalker1991 (changes requested):
Files/Languages/Unofficial/Korean.isl

kristinWalker1991 (commented):
[]()Files/Languages/Unofficial/Korean.isl

PR #423, opened 2022-08-31
Document how to set up code-signing in the CI runs

dscho:
I added support in 9572aae to code-sign build artifacts, but this support was undocumented, though. Let's address that.

While contributing this fix, I noticed that the CI definition was stale and needed to be updated to be able to run successfully again, so I also included this fix in this PR.


martijnlaan:
Thanks!

PR #435, opened 2023-02-15
Update and rename Hindi.islu to Hindi.isl

Zenju:
Hi,

it would be beneficial to have a uniform language file name for integration into scripts. Currently I'm manually converting islu -> isl like for the Hindi example.

Best, Zenju

### DIFF
--- a/Files/Languages/Unofficial/Hindi.isl
+++ b/Files/Languages/Unofficial/Hindi.isl
@@ -13,7 +13,7 @@
 ; understand the '[LangOptions] section' topic in the help file.
 LanguageName=<0939><093F><0902><0926><0940>
 LanguageID=$0439
-LanguageCodePage=0
+LanguageCodePage=65001
 ; If the language you are translating to requires special font faces or
 ; sizes, uncomment any of the following entries and change them accordingly.
 ;DialogFontName=


Zenju:
> Thanks but Hindi.islu is an Inno Setup 5 translation so should remain .islu until it's fully updated.
Okay. FWIW I'm using this translation in Inno 6.2.1 and haven't found any issues or untranslated texts so far.

>  Additionally LanguageCodePage has no effect in UTF8 files so should stay 0 (and also 65001 is not the Hindi codepage).
It's not visible in the git commit, but I've converted the file to UTF8, which is code page 65001. If code page is left at 0, the text is rendered broken/scrambled, but with the previous two changes, it is rendered nicely.

Zenju:
PPS: The Hungarian.isl is missing from the innosetup-6.2.1.exe installer, even though it is marked as an official language: https://jrsoftware.org/files/istrans/#official

Zenju:
> To fix this you should ask the maintainers to update their translation to Inno 6.

No, my translators are adding their translations on top of the Inno Setup translation files, just like in the example I gave above. My script is generating one large file with all translations, expecting a uniform naming convention from Inno as [language name].isl
The Hindi.islu, etc. break this pattern, so I have to manually rename them.

Zenju:
PPPS:
\Languages\Unofficial\Belarusian.isl and
\Languages\Unofficial\Korean.isl

are broken in 6.2.2. They're UTF8-encoded rather than

	LanguageCodePage=949 and
	LanguageCodePage=1251

korean

Zenju:
> > PPPS: \Languages\Unofficial\Belarusian.isl and \Languages\Unofficial\Korean.isl
> > are broken in 6.2.2. They're UTF8-encoded rather than
> 
> Unofficial files are not checked for correctness. As the translation page says they are of unknown quality and you should contact the maintainers to get them fixed.

Why inconvenience your users when you can simply fix an issue yourself in only a few seconds?

Apparently you and I have very different project maintenance philosophies.

hpwamr:
Hello @Zenju ,
I've tried to contact the maintainer of "Hindi.islu"!
For lack of response, I've myself converted the "Hindi.islu" v5.5.3  to "Hindi.isl" v6.2.1+ (some strings are not translated in Hindi!)
It works fine with `encoding=UTF-8 Signature(BOM)` and `LanguageCodePage=0`

Important: it is not enough to convert the file to `"UTF-8"`, but it must be converted to `"UTF-8 Signature(BOM)"`

In attachment my working file: "Hindi.isl" file:  [Hindi.zip](https://github.com/jrsoftware/issrc/files/10831161/Hindi.zip)

PR #437, opened 2023-02-17
Update Korean.isl

VenusGirl:
Most program developers raised the issue of Korean not being displayed properly in the installer, so they changed the encoding to UTF-8 Signature.


VenusGirl (commented):
Fixed

VenusGirl (commented):
Fixed

PR #444, opened 2023-10-26
Update Korean.isl

VenusGirl:
Request to move Korean.isl - Unofficial -> official

Request to move Files/Languages/Unofficial/Korean.isl to Files/Languages so that more program developers can apply Korean smoothly.

The current Korean translation worked perfectly in the installation tests of several programs (Sandboxie, CopyQ, Riso Editor, etc.).


hthetiot:
May be also if you can make your translation right the first time by making it reviewed, instead of keeping changing. No hill intended, just tired of receiving notifications for Korean update for such simple repetitive fixes.

PR #452, opened 2024-03-16
Fix typos in README.md

jogo-:
Removed some trailing white-spaces. 
In chapter#4 "Build Inno Setup", the second part of 1st line, there's a wording issue "which means **you there's** two different build scripts."


jogo- (commented):
Removed the extra "you"

PR #457, opened 2024-04-10
Update German.isl

isaak654:
Source: https://github.com/sandboxie-plus/Sandboxie/issues/1333


isaak654:
> Thanks! Can you also notify Jens Brand please? To avoid this change getting lost if he (as the official maintainer) does an update outside of Git.

@jensbrand

PR #461, opened 2024-06-01
Warn about AppMutex or SetupMutex before languages dialog

brunvonlope:
Resolves: https://groups.google.com/g/innosetup/c/bkRtnHz2ygg

Sorry, I willn't download a whole IDE to code, feels too much. Appreciate help if something is wrong.

### DIFF
--- a/Projects/Src/Main.pas
+++ b/Projects/Src/Main.pas
@@ -3236,6 +3236,19 @@ begin
 
   Log64BitInstallMode;
 
+  { Check if app is running }
+  while CheckForMutexes(ExpandedAppMutex) do
+    if LoggedMsgBox(FmtSetupMessage1(msgSetupAppRunningError, ExpandedAppName),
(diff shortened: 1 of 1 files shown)


martijnlaan:
This doesn't work since it uses initialized variables like ExpandedAppName. Also the message would appear in the default language.

brunvonlope:
> Also the message would appear in the default language.

About the default language, do you mean the system language? If so, this is nothing new since the language dialog itself uses the system default language. Don't understand the point.

> This doesn't work since it uses initialized variables like ExpandedAppName.

Also, since you closed the MR instead of converting it to a draft, I must conclude that it is technically impossible to expanding the variable before the AppMutex dialog.

jordanrussell:
The Select Language dialog is intentionally shown early to maximize the chances that the user will be able to understand any messages that prevent Setup from proceeding.

Yes, the Select Language dialog itself uses default-language text, but that's kind of unavoidable. I mean, we can't just show a combo box with no explanation, and buttons with no text on them.

PR #462, opened 2024-06-02
Fixing a function TNewLinkLabel.AdjustHeight

leserg73:
LM_GETIDEALSIZE message returns Height for a **given Width**.

### DIFF
--- a/Components/BidiCtrls.pas
+++ b/Components/BidiCtrls.pas
@@ -139,7 +139,7 @@ end;
 function TNewLinkLabel.AdjustHeight: Integer;
 begin
   var OldHeight := Height;
-  var IdealHeight := SendMessage(Handle, LM_GETIDEALHEIGHT, 0, 0);
+  var IdealHeight := SendMessage(Handle, LM_GETIDEALSIZE, Width, 0);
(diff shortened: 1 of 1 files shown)


martijnlaan:
But Width is the current width so this should be the same as using GETIDEALHEIGHT? Or are you saying you have seen it fixes the problem which was posted on the forum anyway?

jordanrussell:
Could it be that the `LM_GETIDEALHEIGHT` documentation is just wrong? Wine's implementation simply returns the current font's height, and that looks like what's happening in the forum screen shots as well.

And note that the `LM_GETIDEALSIZE` docs don't say that `lParam` is allowed to be 0.

martijnlaan:
I changed it from LM_GETIDEALSIZE ato LM_GETIDEALHEIGHT in 1d1d6b0854dae4f65414028785e15c1bb5a1e109. Reverted this now since I now see the second doesnt work for me also but the first does. Not sure why and not sure how I missed this back when I did the 'improvement'.

Delphi also has some strange related code I don't understand BTW:

afbeelding

martijnlaan:
Forgot to say, @leserg73 thanks for pointing to the solution!

PR #465, opened 2024-06-11
Place recent files to submenu (test 2)

leserg73:
Added option for clear recent files (under list).

SubRecentFiles


Projects/Src/CompFunc.pas
```diff
@@ -284,6 +285,22 @@ procedure OpenMailingListSite;
...
 
+procedure ClearMRUList(const MRUList: TStringList; const Section, Ident: String);
+var
+  Ini: TConfigIniFile;
+  I: Integer;
+begin
+  Ini := TConfigIniFile.Create;
+  try
+    MRUList.Clear;
+    for I := 0 to MRUListMaxCount-1 do begin
```
martijnlaan, line 296: Isn't there a function like EraseSection or something similar?

Otherwise looks good 👍 

BTW, you should be able to just commit changes to your branch to updatethe PR. So you don't have to close it and create a new one.

leserg73: > Isn't there a function like EraseSection or something similar?

I didn't know it could be done like that. Thanks.

> BTW, you should be able to just commit changes to your branch to update the PR. So you don't have to close it and create a new one.

I'm afraid of breaking something and then having to start over.

martijnlaan:
Did some tweaks and merged it, thanks!

PR #466, opened 2024-06-20
Remove support for Windows Vista in SafeDLLPath.pas

jogo-:
_Support for Windows Vista, Windows Server 2008, and the Itanium architecture removed_


martijnlaan:
Thanks but the Vista notes mean that Vista was the first verson with the problem. I don't think it means that later versions do not have the problem. So this code shouldn't be removed.

jordanrussell:
IMO all of those `SafeLoadLibrary` calls should be removed. (I was going to suggest it for 6.3 but forgot.) They had some value when XP was still supported, but now they're only used on Windows 7 systems that are missing the `SetDefaultDllDirectories` API, which was added in a 2011 update. So it's a code path that benefits practically no one and never receives any testing. (It might crash, for all we know.)

martijnlaan:
Right, that's true 👍

PR #467, opened 2024-06-25
Fix error when logging process output

johnstevenson:
Thank you for introducing the new process output logging features. I think they will be very useful.

However, I noticed that when logging errors from `TCreateProcessOutputReader` the Error param is always False and the FirstLine param is always True. I've fixed this and additionally added an `OutputReader` prefix to the error message so it is clearer in the IDE where the error originated.

I've also unset the `TCreateProcessOutputReader` instance if there was an error when setting up the redirection handles, because the Read method is a no-op in this situation.

I've also made the following changes for you to consider, regarding the handling of errors from the `ExecAndLogOutput` function when `OnLog` is set:

- If there is an error setting up the redirection handles, the program will not be run and the function will return False, with the last error in ResultCode.
- If there is an error reading the program output, the function will return False, with the last error in ResultCode.

(Note that if `OnLog` is not set the program will always run and the function will return True. Likewise when processing a Run/UnistallRun `logoutput` flag, or ISPP's `ExecAndGetFirstLine`).

The logic is that if a user supplies their own function to get the output, then they probably need it so not getting any at all, or only some of it, is an error. This is backed up by `cmd.exe` behaviour which will show an error and not run the command if it cannot set up redirection. For example, running the following as a non-admin:

```
cmd /c echo hello >C:\Windows\System32\output.text
```

returns _Access is denied._ with error code 1.

However, using `LogProcData` in `InstExec` to determine if this is a user-supplied TOnLog procedure could be brittle, considering the number of places it is called from.

### DIFF
--- a/Examples/PowerShell.iss
+++ b/Examples/PowerShell.iss
@@ -59,7 +59,11 @@ end;
 function ExecAndGetFirstLine(const Filename, Params, WorkingDir: String; var ResultCode: Integer): String;
 begin
   Line := '';
-  ExecAndLogOutput(Filename, Params, WorkingDir, SW_SHOWNORMAL, ewWaitUntilTerminated, ResultCode, @ExecAndGetFirstLineLog);
+  try
+    ExecAndLogOutput(Filename, Params, WorkingDir, SW_SHOWNORMAL, ewWaitUntilTerminated, ResultCode, @ExecAndGetFirstLineLog);
+  except
+    Log(GetExceptionMessage);
+  end;
   Result := Line;
 end;
 
--- a/ISHelp/isxfunc.xml
+++ b/ISHelp/isxfunc.xml
@@ -2874,10 +2874,13 @@ end;</pre></example>
         <description><p>Identical to <link topic="isxfunc_Exec">Exec</link> except:</p>
 <p>Console programs are always hidden and the ShowCmd parameter only affects GUI programs, so always using <tt>SW_SHOWNORMAL</tt> instead of <tt>SW_HIDE</tt> is recommended.</p>
 <p>If OnLog is set to <tt>nil</tt> then the output of the executed executable or batch file is logged in Setup's or Uninstall's log file and/or in the Compiler IDE's "Debug Output" view.</p>
-<p>If OnLog is not set to <tt>nil</tt> then the output is sent to the specified function, line by line.</p></description>
+<p>If OnLog is not set to <tt>nil</tt> then the output is sent to the specified function, line by line.</p>
+<p>An exception will be raised if there was an error setting up output redirection.</p></description>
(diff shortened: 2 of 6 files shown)


martijnlaan:
Thanks, looks clean, will take a closer look later.

johnstevenson:
I've actually changed my mind about unsetting the TCreateProcessOutputReader instance if there was an error when setting up the redirection handles.

I didn't apply it to Compile because I missed it, and in retrospect it's probably not worth it considering this that will be a rare event.

jordanrussell:
I wouldn't mind it if the code raised an exception (thus causing the whole operation to fail) if there was an error creating any of the handles. It seems like that should only happen if the system is out of memory or a handle limit has been reached, and in either case, allowing execution to continue will just lead to something else failing later.

jordanrussell:
Hmm, not sure I like the idea of `InstExec` returning False both when the process creation failed, and when the process was successfully started but there was a problem reading the output. How is the caller to differentiate between the two? Currently, `ProcessRunEntry` in `Main.pas` assumes that an `InstExecEx` failure is a `CreateProcess` failure and says that in the error message it displays.

But like handle creation failures, is there any scenario besides possibly "the system is totally out of resources" where `PeekNamedPipe` or `ReadFile` would fail? We shouldn't raise an exception in those cases because it would break out of the loop that waits for the process to terminate. But if you are passing your own output-processing function and want to know about those kinds of failures, can't you already detect them by watching for a call with `Error=True`?

johnstevenson:
Thanks for the feedback. I've implemented your suggestions so that an exception is raised if output redirection fails and `InstExec` does not return False on subsequent read errors.

I agree that these chances of these errors are remote, but there was a spate of errors some time back regarding the nul device driver not being available.

martijnlaan:
Merged this with some minor tweaks, thanks!

jordanrussell:
> I agree that these chances of these errors are remote, but there was a spate of errors some time back regarding the nul device driver not being available.

Can you expand on that? What caused the NUL device to be unavailable, and is it still an issue today?

johnstevenson:
Searching https://answers.microsoft.com/en-us for "null device driver" brings up issues from 2016 to 2022 where it is missing or broken. These lead to a rabbit hole of solutions and possible fixes but no clear reason why it happened.

For Composer (PHP Package manager) we even created a Troubleshooting FAQ for it:
https://getcomposer.org/doc/articles/troubleshooting.md#proc-open-failed-to-open-stream-errors-windows-

Git For Windows also had issues: https://github.com/git-for-windows/git/issues/1554

For PHP the issue was also caused by some versions not recognizing the reparse points semantics of the OneDrive file system (when running code from there).

It happened to me and was nothing to do with OneDrive (which we didn't know about then). The problem disappeared after a restart. Perhaps it was caused and solved by automatic Windows Updates (of which there were several around at that time), but I cannot be certain.

PR #468, opened 2024-06-28
Proposing an ExecAndCaptureOutput function

johnstevenson:
Following on from ExecAndLogOutput, this is a proposal for an `ExecAndCaptureOutput` function that captures program output from the stdout and stderr streams and writes it to a TExecOutput record.

```delphi
function ExecAndCaptureOutput(const Filename, Params, WorkingDir: String; const ShowCmd: Integer;
  const Wait: TExecWait; var ResultCode: Integer; var Output: TExecOutput): Boolean;
    
TExecOutput = record
  StdOut: TArrayOfString;
  StdErr: TArrayOfString;
end;
```
    
This is different from ExecAndLogOutput which merges the streams. The stderr stream is often used for non-error but informative output, so to split the streams you need to use Exec with cmd.exe to redirect to temporary files.  This new function does away with all that plumbing.

The PR is split into two commits. The first simplifies the exec logging parameters by wrapping them in a `TOutputParams` record, which lays the foundations for the second, which implements the new function.

The output is sent to files in the `_isetup` sub-directory of the temp folder. I don't know if this is a suitable location. The files are created with `CREATE_ALWAYS` in case en exception is raised before they are deleted.

### DIFF
--- a/ISHelp/isxfunc.xml
+++ b/ISHelp/isxfunc.xml
@@ -2875,7 +2875,7 @@ end;</pre></example>
         <description><p>Identical to <link topic="isxfunc_Exec">Exec</link> except:</p>
 <p>Program output from the stdout and stderr streams is captured. This is different from <link topic="isxfunc_ExecAndLogOutput">ExecAndLogOutput</link> which merges the streams.</p>
 <p>Console programs are always hidden and the ShowCmd parameter only affects GUI programs, so always using <tt>SW_SHOWNORMAL</tt> instead of <tt>SW_HIDE</tt> is recommended.</p>
-<p>An exception will be raised if there was an error setting up output redirection (which should be very rare). The error is logged in Setup's or Uninstall's log file and/or in the Compiler IDE's "Debug Output" view.</p></description>
+<p>An exception will be raised if there was an error setting up output redirection (which should be very rare).</p></description>
         <remarks><p>Parameter <tt>Wait</tt> must always be set to <tt>ewWaitUntilTerminated</tt> when calling this function.</p>
 <p>TExecOutput is defined as:</p>
 <pre>
@@ -2918,7 +2918,7 @@ end;</pre></example>
 <p>TOnLog is defined as:</p>
 <p><tt>TOnLog = procedure(const S: String; const Error, FirstLine: Boolean);</tt></p>
 <p>Parameter S is the output line when Error is False, otherwise an error message. FirstLine is True if this is the first line of output from the program, otherwise False.</p>
-<p>Error will be True if setting up output redirection or reading the output failed, or if the output size exceeded 10mb. There is no further output after an error.</p></remarks>
+<p>Error will be True if reading the output failed (which should be very rare), or if the output size exceeded 10mb. There is no further output after an error.</p></remarks>
         <example><pre>var
   Line: String;
 
--- a/ISHelp/isxfunc.xml
+++ b/ISHelp/isxfunc.xml
@@ -1651,7 +1651,8 @@ begin
     // handle failure if necessary; ResultCode contains the error code
   end;
 end;</pre></example>
-        <seealso><p><link topic="isxfunc_ExecAndLogOutput">ExecAndLogOutput</link><br />
+        <seealso><p><link topic="isxfunc_ExecAndCaptureOutput">ExecAndCaptureOutput</link><br />
+<link topic="isxfunc_ExecAndLogOutput">ExecAndLogOutput</link><br />
 <link topic="isxfunc_ExecAsOriginalUser">ExecAsOriginalUser</link></p></seealso>
       </function>
       <function>
@@ -2868,6 +2869,43 @@ end;</pre></example>
         <description><p>Logs the specified string in Setup's or Uninstall's log file and/or in the Compiler IDE's "Debug Output" view.</p></description>
         <remarks><p>Calls to this function are ignored if logging is not enabled (via the <link topic="setupcmdline">/LOG</link> command line parameter or the <link topic="setup_setuplogging">SetupLogging</link> [Setup] section directive or the <link topic="setup_uninstalllogging">UninstallLogging</link> [Setup] section directive or debugging from the Compiler IDE).</p></remarks>
       </function>
+<function>
+        <name>ExecAndCaptureOutput</name>
+        <prototype>function ExecAndCaptureOutput(const Filename, Params, WorkingDir: String; const ShowCmd: Integer; const Wait: TExecWait; var ResultCode: Integer; var Output: TExecOutput): Boolean;</prototype>
+        <description><p>Identical to <link topic="isxfunc_Exec">Exec</link> except:</p>
+<p>Program output from the stdout and stderr streams is captured. This is different from <link topic="isxfunc_ExecAndLogOutput">ExecAndLogOutput</link> which merges the streams.</p>
+<p>Console programs are always hidden and the ShowCmd parameter only affects GUI programs, so always using <tt>SW_SHOWNORMAL</tt> instead of <tt>SW_HIDE</tt> is recommended.</p>
+<p>An exception will be raised if there was an error setting up output redirection (which should be very rare). The error is logged in Setup's or Uninstall's log file and/or in the Compiler IDE's "Debug Output" view.</p></description>
+        <remarks><p>Parameter <tt>Wait</tt> must always be set to <tt>ewWaitUntilTerminated</tt> when calling this function.</p>
+<p>TExecOutput is defined as:</p>
+<pre>
+  TExecOutput = record
+    StdOut: TArrayOfString;
+    StdErr: TArrayOfString;
+    Error: Boolean;
+  end;
+</pre>
+<p>Error will be True if an exception was raised or if there was an error reading the output (which should be very rare) or if the total output size exceeded 10mb. The error is logged in Setup's or Uninstall's log file and/or in the Compiler IDE's "Debug Output" view. There is no further output after an error.</p></remarks>
+        <example><pre>var
+  ResultCode: Integer;
+  Output: TExecOutput;
+
+begin
+  try
+    // Get the system configuration
+    Result := ExecAndCaptureOutput(ExpandConst('{cmd}', 'systeminfo', '', SW_SHOWNORMAL, ewWaitUntilTerminated, ResultCode, Output);
+  except
+    Result := False
+    Log(GetExceptionMessage);
+  end;
+
+  if Result then begin
+    // Do something with Output.StdOut
+  end;
+end;</pre></example>
+        <seealso><p><link topic="isxfunc_Exec">Exec</link><br />
+<link topic="isxfunc_ExecAndLogOutput">ExecAndLogOutput</link></p></seealso>
+      </function>
       <function>
         <name>ExecAndLogOutput</name>
         <prototype>function ExecAndLogOutput(const Filename, Params, WorkingDir: String; const ShowCmd: Integer; const Wait: TExecWait; var ResultCode: Integer; const OnLog: TOnLog): Boolean;</prototype>
@@ -2903,7 +2941,8 @@ begin
   end;
   Result := Line;
 end;</pre></example>
-        <seealso><p><link topic="isxfunc_Exec">Exec</link></p></seealso>
+        <seealso><p><link topic="isxfunc_Exec">Exec</link><br />
+<link topic="isxfunc_ExecAndCaptureOutput">ExecAndCaptureOutput</link></p></seealso>
       </function>
     </subcategory>
   </category>
--- a/ISHelp/isxfunc.xml
+++ b/ISHelp/isxfunc.xml
@@ -2885,7 +2885,8 @@ end;</pre></example>
     Error: Boolean;
   end;
 </pre>
-<p>Error will be True if there was an error reading the output (which should be very rare) or if the total output size exceeded 10mb. The error is logged in Setup's or Uninstall's log file and/or in the Compiler IDE's "Debug Output" view. There is no further output after an error.</p></remarks>
+<p>Error will be True if there was an error reading the output (which should be very rare) or if the output is too big. The error is logged in Setup's or Uninstall's log file and/or in the Compiler IDE's "Debug Output" view. There is no further output after an error.</p>
+<p>Output is limited to a total size of 10 million bytes or a maximum of 1 million lines.</p></remarks>
         <example><pre>var
   ResultCode: Integer;
   Output: TExecOutput;
@@ -2918,7 +2919,8 @@ end;</pre></example>
 <p>TOnLog is defined as:</p>
 <p><tt>TOnLog = procedure(const S: String; const Error, FirstLine: Boolean);</tt></p>
 <p>Parameter S is the output line when Error is False, otherwise an error message. FirstLine is True if this is the first line of output from the program, otherwise False.</p>
-<p>Error will be True if reading the output failed (which should be very rare), or if the output size exceeded 10mb. There is no further output after an error.</p></remarks>
+<p>Error will be True if reading the output failed (which should be very rare), or if the output is too big. There is no further output after an error.</p>
+<p>Output is limited to a total size of 10 million bytes or a maximum of 1 million lines.</p></remarks>
         <example><pre>var
   Line: String;
 
--- a/ISHelp/isxfunc.xml
+++ b/ISHelp/isxfunc.xml
@@ -2885,7 +2885,7 @@ end;</pre></example>
     Error: Boolean;
   end;
 </pre>
-<p>Error will be True if an exception was raised or if there was an error reading the output (which should be very rare) or if the total output size exceeded 10mb. The error is logged in Setup's or Uninstall's log file and/or in the Compiler IDE's "Debug Output" view. There is no further output after an error.</p></remarks>
+<p>Error will be True if there was an error reading the output (which should be very rare) or if the total output size exceeded 10mb. The error is logged in Setup's or Uninstall's log file and/or in the Compiler IDE's "Debug Output" view. There is no further output after an error.</p></remarks>
         <example><pre>var
   ResultCode: Integer;
   Output: TExecOutput;
--- a/Projects/Src/CmnFunc2.pas
+++ b/Projects/Src/CmnFunc2.pas
@@ -36,21 +36,6 @@ type
   TLogProc = procedure(const S: String; const Error, FirstLine: Boolean; const Data: NativeInt);
   TOutputMode = (omLog, omCapture);
 
-  TOutputParams = record
-    Enabled: Boolean;
-    Mode: TOutputMode;
-    LogProc: TLogProc;
-    LogProcData: NativeInt;
-    OutList: TStringList;
-    ErrList: TStringList;
-    Error: Boolean;
-    class operator Initialize(out Dest: TOutputParams);
-    class operator Finalize (var Dest: TOutputParams);
-    class function WithLogData(const ALogProc: TLogProc; ALogProcData: NativeInt): TOutputParams; static;
-    procedure SetLogData(const ALogProc: TLogProc; ALogProcData: NativeInt);
-    procedure SetCaptureData(const ALogProc: TLogProc);
-  end;
-
   TCreateProcessOutputReader = class
   private
     FOKToRead: Boolean;
@@ -68,18 +53,24 @@ type
     FNextLineIsFirstLine: Boolean;
     FLastLogErrorMessage: String;
     FMode: TOutputMode;
-    FOutputParams: ^TOutputParams;
+    FCaptureOutList: TStringList;
+    FCaptureErrList: TStringList;
+    FCaptureError: Boolean;
     procedure CloseAndClearHandle(var Handle: THandle);
     procedure LogErrorFmt(const S: String; const Args: array of const);
-    function PipeCreate(var Read, Write: THandle; SecurityAttr: TSecurityAttributes): Boolean;
     procedure DoRead(var PipeRead: THandle; var Buffer: AnsiString; const LastRead: Boolean);
+    function GetCaptureOutList: TStringList;
+    function GetCaptureErrList: TStringList;
   public
-    constructor Create(const OutputParams: TOutputParams);
+    constructor Create(const ALogProc: TLogProc; const ALogProcData: NativeInt; AMode: TOutputMode = omLog);
     destructor Destroy; override;
     procedure UpdateStartupInfo(var StartupInfo: TStartupInfo);
     procedure NotifyCreateProcessDone;
     procedure Read(const LastRead: Boolean);
     property MaxTotalBytesToRead: Cardinal read FMaxTotalBytesToRead write FMaxTotalBytesToRead;
+    property CaptureOutList: TStringList read GetCaptureOutList;
+    property CaptureErrList: TStringList read GetCaptureErrList;
+    property CaptureError: Boolean read FCaptureError;
   end;
 
   TRegView = (rvDefault, rv32Bit, rv64Bit);
@@ -1617,79 +1608,34 @@ begin
     Result := 0;
 end;
 
-{ TOutputParams }
-
-class operator TOutputParams.Initialize(out Dest: TOutputParams);
-begin
-  Dest.Enabled := False;
-  Dest.Mode := omLog;
-  Dest.LogProc := nil;
-  Dest.LogProcData := 0;
-  Dest.OutList := nil;
-  Dest.ErrList := nil;
-  Dest.Error := False;
-end;
-
-class operator TOutputParams.Finalize (var Dest: TOutputParams);
-begin
-  if Dest.OutList <> nil then
-    Dest.OutList.Free;
-
-  if Dest.ErrList <> nil then
-    Dest.ErrList.Free;
-end;
-
-class function TOutputParams.WithLogData(const ALogProc: TLogProc;
-  ALogProcData: NativeInt): TOutputParams;
-begin
-  Result.SetLogData(ALogProc, ALogProcData);
-end;
-
-procedure TOutputParams.SetLogData(const ALogProc: TLogProc; ALogProcData: NativeInt);
-begin
-  if not Assigned(ALogProc) then
-    raise Exception.Create('ALogProc is required');
-
-  Enabled := True;
-  Mode := omLog;
-  LogProc := ALogProc;
-  LogProcData := ALogProcData;
-end;
-
-procedure TOutputParams.SetCaptureData(const ALogProc: TLogProc);
-begin
-  if not Assigned(ALogProc) then
-    raise Exception.Create('ALogProc is required');
-
-  if Assigned(OutList) or Assigned(ErrList) then
-    raise Exception.Create('Output lists already assigned');
-
-  Enabled := True;
-  Mode := omCapture;
-  LogProc := ALogProc;
-  LogProcData := 0;
-  OutList := TStringList.Create;
-  ErrList := TStringList.Create;
-  Error := False;
-end;
-
 { TCreateProcessOutputReader }
 
-constructor TCreateProcessOutputReader.Create(const OutputParams: TOutputParams);
-begin
-  if not Assigned(OutputParams.LogProc) then
-    raise Exception.Create('LogProc is required');
+constructor TCreateProcessOutputReader.Create(const ALogProc: TLogProc;
+  const ALogProcData: NativeInt; AMode: TOutputMode = omLog);
 
-  if OutputParams.Mode = omCapture then begin
-    if (not Assigned(OutputParams.OutList) or not Assigned(OutputParams.ErrList)) then
-      raise Exception.Create('OutList and ErrList are required');
-
-    FOutputParams := @OutputParams;
+  function PipeCreate(var Read, Write: THandle; SecurityAttr: TSecurityAttributes): Boolean;
+  begin
+    Result := False;
+    if not CreatePipe(Read, Write, @SecurityAttr, 0) then
+      LogErrorFmt('CreatePipe failed (%d).', [GetLastError])
+    else if not SetHandleInformation(Read, HANDLE_FLAG_INHERIT, 0) then
+      LogErrorFmt('SetHandleInformation failed (%d).', [GetLastError])
+    else
+      Result := True;
   end;
 
-  FMode := OutputParams.Mode;
-  FLogProc := OutputParams.LogProc;
-  FLogProcData := OutputParams.LogProcData;
+begin
+  if not Assigned(ALogProc) then
+    raise Exception.Create('ALogProc is required');
+
+  if AMode = omCapture then begin
+    FCaptureOutList := TStringList.Create;
+    FCaptureErrList := TStringList.Create;
+  end;
+
+  FMode := AMode;
+  FLogProc := ALogProc;
+  FLogProcData := ALogProcData;
   FNextLineIsFirstLine := True;
 
   var SecurityAttributes: TSecurityAttributes;
@@ -1731,6 +1677,8 @@ begin
   CloseAndClearHandle(FStdOutPipeWrite);
   CloseAndClearHandle(FStdErrPipeRead);
   CloseAndClearHandle(FStdErrPipeWrite);
+  FCaptureOutList.Free;
+  FCaptureErrList.Free;
   inherited;
 end;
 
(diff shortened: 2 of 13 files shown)


martijnlaan:
I will provide some review comments tomorrow, but I want to say that I'm very impressed with the quality and completeness of your pull requests. Thanks! 👍

Projects/Src/CmnFunc2.pas
```diff
@@ -137,6 +153,8 @@ procedure WaitMessageWithTimeout(const Milliseconds: DWORD);
 function MoveFileReplace(const ExistingFileName, NewFileName: String): Boolean;
 procedure TryEnableAutoCompleteFileSystem(Wnd: HWND);
 procedure CreateMutex(const MutexName: String);
+function GetLogOutputParams(const LogProc: TLogProc; const LogProcData: NativeInt): TOutputParams;
```
martijnlaan, line 156: Can you make these member functions of TOutputParams instead?

Projects/Src/CmnFunc2.pas
```diff
@@ -1547,6 +1565,32 @@ procedure CreateMutex(const MutexName: String);
...
+function GetCaptureOutputParams(const TempDir: String): TOutputParams;
+begin
+  if TempDir = '' then
+    raise Exception.Create('TempDir is required');
+
+  Result.Log := True;
+  Result.Mode := omCapture;
+  Result.LogProc := nil;
+  Result.LogProcData := 0;
+  Result.OutFile := AddBackslash(TempDir) + 'stdout.txt';
```
martijnlaan, line 1590: Perhaps a little more unique name? Like _stdout.txt at least.

Projects/Src/CmnFunc2.pas
```diff
@@ -1618,17 +1666,31 @@ constructor TCreateProcessOutputReader.Create(const ALogProc: TLogProc;
...
+      if not CreateOutputFile(OutputParams.OutFile, SecurityAttributes, Outfile) then
+        LogErrorFmt('CreateOutputFile failed (%d).', [GetLastError])
+      else if not CreateOutputFile(OutputParams.ErrFile, SecurityAttributes, ErrFile) then
+        LogErrorFmt('CreateOutputFile failed (%d).', [GetLastError])
+      else begin
+        FStdOutFile := OutFile;
+        FStdErrFile := ErrFile;
 
-      FOKToRead := True;
-      FMaxTotalBytesToRead := 10*1024*1024;
```
martijnlaan, line 1631: I think there still should always be a limit. Without it a rogue child process can cause us to fill up the disk?

Projects/Src/CmnFunc2.pas
```diff
@@ -1618,17 +1666,31 @@ constructor TCreateProcessOutputReader.Create(const ALogProc: TLogProc;
...
+        FOKToRead := True;
```
martijnlaan, line 1692: Perhaps this var needs a rename. Now it suggests there will always be reading but this is not true.

Projects/Src/InstFunc.pas
```diff
@@ -951,8 +953,8 @@ function InstExec(const DisableFsRedir: Boolean; const Filename, Params: String;
     var InheritHandles := False;
     var dwCreationFlags: DWORD := CREATE_DEFAULT_ERROR_MODE;
 
-    if Log and Assigned(LogProc) and (Wait = ewWaitUntilTerminated) then begin
-      OutputReader := TCreateProcessOutputReader.Create(LogProc, LogProcData);
+    if OutputParams.Log and (Wait = ewWaitUntilTerminated) then begin
```
martijnlaan, line 956: Why doesn't this check LogProc anymore?

Projects/Src/Main.pas
```diff
@@ -3921,10 +3921,12 @@ procedure ProcessRunEntry(const RunEntry: PSetupRunEntry);
...
       if not(roSkipIfDoesntExist in RunEntry.Options) or
          NewFileExistsRedir(DisableFsRedir, ExpandedFilename) then begin
+        var OutputParams: TOutputParams;
+        if GetLogActive and (roLogOutput in RunEntry.Options) then
+          OutputParams := GetLogOutputParams(RunExecLog, 0);
         if not InstExecEx(RunAsOriginalUser, DisableFsRedir, ExpandedFilename,
            ExpandedParameters, ExpandConst(RunEntry.WorkingDir),
-           Wait, RunEntry.ShowCmd, ProcessMessagesProc, GetLogActive and (roLogOutput in RunEntry.Options),
-           RunExecLog, 0, ErrorCode) then
+           Wait, RunEntry.ShowCmd, ProcessMessagesProc, OutputParams, ErrorCode) then
```
martijnlaan, line 3929: OutputParams doesnt have a constructor so now InstExec can receive unitialized data?

Same is happening in other places.

Projects/Src/ScriptFunc_R.pas
```diff
@@ -808,12 +808,40 @@ procedure ExecAndLogOutputLog(const S: String; const Error, FirstLine: Boolean;
...
+  end;
+
 procedure ExecAndLogOutputLogCustom(const S: String; const Error, FirstLine: Boolean; const Data: NativeInt);
 begin
   var OnLog := TOnLog(PMethod(Data)^);
   OnLog(S, Error, FirstLine);
 end;
 
+{ Adapted from LoadStringsFromFile. Raises an exception on failure. }
+procedure ExecAndCaptureOutputLog(const FileName: String; Arr: PPSVariantIFC);
```
martijnlaan, line 823: Function name seems confusing. Has Log in its name but is not related to logging? It's not even really specific to output reading but just a generic file to strings. Why doesn't the caller use LoadStringsFromFile?

martijnlaan (changes requested):

Projects/Src/CmnFunc2.pas
johnstevenson: Will do.

Projects/Src/CmnFunc2.pas
johnstevenson: Indeed. I'll use your suggestion unless you would prefer something else?

Projects/Src/CmnFunc2.pas
johnstevenson: FMaxTotalBytesToRead is still set to this value if `FMode = omLog`.

Are you saying that if `FMode = omCapture` then we should also apply the restriction? At the moment the file handles are just passed to the child process for it to do the output writing.

> Without it a rogue child process can cause us to fill up the disk?

A rogue child process can do anything really. It could fill the disk without output redirection setup. Is this not the responsibility of the user, who chooses the process to run?

Projects/Src/CmnFunc2.pas
johnstevenson: I can re-introduce `FCreatedHandles` to make things more explicit.

Projects/Src/InstFunc.pas
johnstevenson: Because when capturing output there is no LogProc.

Projects/Src/Main.pas
johnstevenson: Sorry, I wrongly presumed the variable would be initialized. I will fix this.

Projects/Src/ScriptFunc_R.pas
johnstevenson: Yes, I struggled with the name and was trying to make it fit with the others.

It was my intention to use LoadStringsFromFile but it is a nested function. If I could use it, then I would also need to raise an exception if it returned false and also delete the file, for both files.

This is quite a lot of boilerplate to add to the flow after `Stack.SetInt(PStart-6, ResultCode);` so I thought a separate procedure would be clearer. But I'll do whatever you prefer.

Perhaps `ExecCaptureFinalize` would be a better name?

Projects/Src/CmnFunc2.pas
martijnlaan: Let's do _isstdout.txt and _isstderr.txt, this is most consistent with other things it places there ike _iscrypt.dll.

Projects/Src/CmnFunc2.pas
martijnlaan: Leave it as is for now, I'll see if I'll rename it later.

Projects/Src/ScriptFunc_R.pas
martijnlaan: Yes that's better. I'll see if I'll rename it some more later.

Projects/Src/CmnFunc2.pas
martijnlaan: Yes please also apply the restriction on omCapture. The limit is there to prevent setup.log from filling the disk so similarly applies to these txt's. I don't mind if if you grow the limit some if you think 10 MB isnt enough even though it seems like quite a lot.

Projects/Src/Compile.pas
```diff
@@ -7387,7 +7388,8 @@ procedure TSetupCompiler.SignCommand(const AName, ACommand, AParams, AExeFilenam
     StartupInfo.dwFlags := STARTF_USESHOWWINDOW;
     StartupInfo.wShowWindow := IfThen(RunMinimized, SW_SHOWMINNOACTIVE, SW_SHOWNORMAL);
 
-    var OutputReader := TCreateProcessOutputReader.Create(SignCommandLog, NativeInt(Self));
+    OutputParams := GetLogOutputParams(SignCommandLog, NativeInt(Self));
```
martijnlaan, line 7391: Please use inline variables always in new code. You may also make all the other variables (in the same touched function, not everywhere else) inline if you want but it's okay if you don't.

So 'var OutputParams'.

Projects/Src/ScriptFunc_R.pas
```diff
@@ -808,12 +808,40 @@ procedure ExecAndLogOutputLog(const S: String; const Error, FirstLine: Boolean;
...
+var
+  F: TTextFileReader;
```
martijnlaan, line 825: Same note about using inline variables.

Projects/Src/CmnFunc2.pas
johnstevenson: Okay, so to clarify omCapture, I would need to write a method to read from the pipe and write to the file, and call it for the stdout and stderr streams?

martijnlaan: Right, you can enforce it otherwise. Doing that seems a bit over the top. Please ignore my comment.

Projects/Src/Compile.pas
johnstevenson: Will do

Projects/Src/ScriptFunc_R.pas
johnstevenson: Will do

johnstevenson:
I think I've covered everything. I squashed it because the first commit message no longer applied due to the changes.

`TOutputParams` now has an initializer. I renamed the `Log` property to `Enabled` for clarity (I hope).

johnstevenson:
Actually, please hang fire. I missed setting the LogProc for error messages when Mode is omCapture.

jordanrussell:
It's not clear to me why the output is being redirected to files rather than reusing (some of) the existing pipe code. Is it just because `TTextFileReader` only accepts a file name?

Some potential issues with redirecting to files come to mind:

- There's no way (that I know of) to limit the maximum size of a file. If a process gets stuck in an infinite loop writing to stdout/stderr, then given enough time, it will eventually fill up the whole disk. Setup may also run out of address space trying to read in an unusually large file (though that could be mitigated by checking the file size before reading).
- The file deletion will fail in this scenario:
  1. Setup spawns process B with stdout/stderr redirected to files.
  2. Process B spawns process C, which inherits process B's standard handles.
  3. Process B exits, but without waiting for process C to terminate (perhaps by design).
  4. Setup sees process B has terminated, and then tries to delete the files. This fails because process C still holds handles to them.
- When a process creates a file, AV software may initiate a scan on that file after the process closes the handle. If the process tries to reopen the same file with write access before the AV scan has completed, it may fail with an error like "access denied". This has impacted Inno Setup's compiler for ages, and the usual advice given is to exclude the output directory from scanning. I'm a little concerned a similar problem could occur here if the .txt files are created/deleted/recreated "too fast".

It seems this could be implemented with pipes, with some deviations from the existing pipe code:
- Create two pipes instead of one (obviously)
- Instead of trying to break the pipe input into lines and feeding those lines to `LogProc`, forward all bytes read from the pipes into a pair of `TMemoryStream` or similar. (Stop if the streams get excessively large.)
- Pass streams instead of file names to `ExecAndCaptureFinalize`

Now the problem: `TTextFileReader` only supports reading from files (and changing that would require major re-engineering). As an alternative, `TStringList` could be used, with a stream passed to the `LoadFromStream` method.

johnstevenson:
The reason for redirecting to files was to avoid worrying about the size of the output and let the child process directly write to the files. However, I clearly hadn't considered all the pitfalls, so many thanks for the explanation.

I was actually working on an implementation along the that lines you have suggested but using TStringList (adding the parsed lines). In `ExecAndCaptureFinalize` I was intending to set the TArrayOfStrings length to the Count property, then iterate and copy the items. Does that sound reasonable?

What would you consider to be "excessively large" output? As Martijn pointed out above, 10mb does seem quite a lot, but there are 2 streams in this scenario.

jordanrussell:
> I was actually working on an implementation along the that lines you have suggested but using TStringList (adding the parsed lines). In `ExecAndCaptureFinalize` I was intending to set the TArrayOfStrings length to the Count property, then iterate and copy the items. Does that sound reasonable?

Yes, that makes perfect sense. I'm not sure what I was thinking when I suggested the intermediate `TMemoryStream`.

> What would you consider to be "excessively large" output? As Martijn pointed out above, 10mb does seem quite a lot, but there are 2 streams in this scenario.

Oh - I totally missed that thread due to it being marked "resolved" and initially collapsed...

The situation here is a little different as all lines are kept in memory at once. In the worst case, 10MB of output may cause 5M strings to be allocated on the heap, if every line contains 1 character + 1 line ending character (just LF or just CR). IIRC, a 1-character UnicodeString requires a 16-byte heap allocation (2 bytes for the character + 2 bytes for the null terminator + 12 bytes for the string type's internal fields). Then there's some additional overhead from the memory manager; I have no idea what it is, but let's say it's another 16 bytes. Then in the `TArrayOfString`, 4 bytes are needed to hold a pointer to the string (assuming `TArrayOfString` doesn't maintain any other data for the elements; I haven't checked).

So, 5M * (16 + 16 + 4) = 180MB memory required. If stdout and stderr *both* contain 5M lines, then double that to 360MB.

Is that too much for the worst case? I guess a better question is, what's the most output a non-malfunctioning program should reasonably be expected to produce? I'm perfectly fine with lowering the 10MB limit. I just picked that number because it sounded like more than anyone would realistically need.

BTW, there may be a performance problem with reading a full 10MB out of the pipes. In the `CreatePipe` call, 0 is being passed for the buffer size, which means "the default buffer size". I don't know what that is. If it's 64KB, then it would take at minimum 160 pipe reads to get all 10MB. With a sleep of 50ms between reads, that adds up to 8000ms. We may want to consider reducing the sleep time to 10ms or so, and/or specifying an explicit buffer size like 256KB or 1MB.

johnstevenson:
The latest 46be71d implements a pipe rather than a file based solution.

I haven't addressed the maximum output size, because I have no idea what "the most output a non-malfunctioning program should reasonably be expected to produce". But considering this is console text output I would have though that something like 2mb (for both streams) would be more than adequate?

Projects/Src/CmnFunc2.pas
```diff
@@ -1594,17 +1617,80 @@ function TOneShotTimer.TimeRemaining: Cardinal;
...
+constructor TCreateProcessOutputReader.Create(const OutputParams: TOutputParams);
+begin
+  if not Assigned(OutputParams.LogProc) then
+    raise Exception.Create('LogProc is required');
+
+  if OutputParams.Mode = omCapture then begin
+    if (not Assigned(OutputParams.OutList) or not Assigned(OutputParams.ErrList)) then
+      raise Exception.Create('OutList and ErrList are required');
+
+    FOutputParams := @OutputParams;
```
jordanrussell, line 1687: Stashing a pointer to a `const` argument, and then dereferencing it after the function returns is rather disturbing! If the caller were to accidentally allow `OutputParams` to go out of scope (and be finalized) then there could be a use-after-free problem.

There might also be a bad outcome if `OutputParams` were passed to a function in a non-`const` argument. That would cause a copy of the record to be made (I think). When the copy is finalized, `OutList` and `ErrList` would be freed, and then when the original record is finalized, they would be freed again -- so, a double-free problem.

IMO, it would be safer and ultimately improve readability if `TOutputParams` were just a regular class that has to be explicitly created and freed. AFAICS `TOutputParams.Initialize` wouldn't be needed at all since classes are implicitly zero-initialized. And with a class, `InstExec` could accept `OutputParams = nil` to mean no output reader.

Projects/Src/CmnFunc2.pas
```diff
@@ -1653,6 +1746,20 @@ procedure TCreateProcessOutputReader.LogErrorFmt(const S: String; const Args: ar
...
+
+function TCreateProcessOutputReader.PipeCreate(var Read, Write: THandle;
+  SecurityAttr: TSecurityAttributes): Boolean;
+begin
+  if not CreatePipe(Read, Write, @SecurityAttr, 0) then
+    LogErrorFmt('CreatePipe failed (%d).', [GetLastError])
+  else if not SetHandleInformation(Read, HANDLE_FLAG_INHERIT, 0) then
+    LogErrorFmt('SetHandleInformation failed (%d).', [GetLastError]);
+
+  Result := GetLastError = 0;
```
jordanrussell, line 1762: As noted in the `GetLastError` docs, when functions succeed, it is usually undefined whether or not they reset the last error code to 0. So, only the `BOOL` return values should be trusted.

Also, we usually use nested functions when a function is only used internally by a single function.

jordanrussell:
> I haven't addressed the maximum output size, because I have no idea what "the most output a non-malfunctioning program should reasonably be expected to produce". But considering this is console text output I would have though that something like 2mb (for both streams) would be more than adequate?

Yeah, 2MB does sound plenty adequate for typical console utilities. What I'm not sure about are atypical uses where someone might want to get a dump of data, as opposed to collecting status or error messages for debugging purposes. I guess they could always fall back to redirecting to a file using `cmd.exe`, or use an external program that is better equipped for data processing.

Projects/Src/CmnFunc2.pas
johnstevenson: Ah, I should know that! Will refactor as a nested function.

ISHelp/isxfunc.xml
```diff
@@ -2868,6 +2869,43 @@ end;</pre></example>
...
+        <remarks><p>Parameter <tt>Wait</tt> must always be set to <tt>ewWaitUntilTerminated</tt> when calling this function.</p>
+<p>TExecOutput is defined as:</p>
+<pre>
+  TExecOutput = record
+    StdOut: TArrayOfString;
+    StdErr: TArrayOfString;
+    Error: Boolean;
+  end;
+</pre>
+<p>Error will be True if an exception was raised or if there was an error reading the output (which should be very rare) or if the total output size exceeded 10mb. The error is logged in Setup's or Uninstall's log file and/or in the Compiler IDE's "Debug Output" view. There is no further output after an error.</p></remarks>
```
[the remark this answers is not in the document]
johnstevenson: The Error field is initialized to True before the function is called.

I only did this because when testing various failure scenarios it seemed odd (to me) that Error was False after an exception was raised and caught in userland. But I suppose the user could set it to True themselves when catching it.

I'll fix this.

Projects/Src/CmnFunc2.pas
```diff
@@ -1594,17 +1617,80 @@ function TOneShotTimer.TimeRemaining: Cardinal;
...
+  Dest.LogProc := nil;
+  Dest.LogProcData := 0;
+  Dest.OutList := nil;
+  Dest.ErrList := nil;
+  Dest.Error := False;
+end;
+
+class operator TOutputParams.Finalize (var Dest: TOutputParams);
+begin
+  if Dest.OutList <> nil then
```
martijnlaan, line 1635: nil checks are not needed before calling Free, Delphi already does that automatically. From the docs: "Unlike Destroy, Free is successful even if the object is nil;".

Projects/Src/CmnFunc2.pas
johnstevenson: Sorry, it is pretty shocking. I was trying to avoid creating another class with its associated create and free calls all over the place.

Would it be acceptable to pass OutputParams to OutputReader.Read instead?

Projects/Src/CmnFunc2.pas
johnstevenson: Ah, okay. Will fix if we do not change this to a class.

Projects/Src/CmnFunc2.pas
```diff
@@ -1689,14 +1809,19 @@ procedure TCreateProcessOutputReader.Read(const LastRead: Boolean);
 
   procedure LogLine(const S: AnsiString);
   begin
-    FLogProc(UTF8ToString(S), False, FNextLineIsFirstLine, FLogProcData);
-    FNextLineIsFirstLine := False;
+    if FMode = omLog then begin
+      FLogProc(UTF8ToString(S), False, FNextLineIsFirstLine, FLogProcData);
+      FNextLineIsFirstLine := False;
+    end else if PipeRead = FStdOutPipeRead then
+      FOutputParams^.OutList.Add(UTF8ToString(S))
```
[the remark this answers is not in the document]
johnstevenson: Okay

Projects/Src/CmnFunc2.pas
johnstevenson: In retrospect I think it would be better to dispense with `TOutputParams` altogether and supply a `TCreateProcessOutputReader` (which is a bit of a mouthful) instance or `nil` to the `InstExec` etc procedures.

martijnlaan: I havent studied the latest version in detail but was getting this feeling as well so I think you're right.

johnstevenson: Thanks. I think it has worked out okay: f35910f137416564934b89f9c06b7f670da4af99

jordanrussell:
👍 So much cleaner!

One thing I'm not quite understanding, and this was in the previous PR too, is why, when `CreatePipe` or `SetHandleInformation` fail, the exception isn't raised immediately, but is instead delayed until `UpdateStartupInfo` is called.

Not sure if you're aware, but in Delphi, when a constructor raises an exception, the destructor gets called automatically before the exception is propagated out. So any handles that were already created before the exception was raised will be closed properly.

Also why does it log the error *and* raise an exception? I believe in most cases that will result in the error appearing in the log twice. Just raising the exception seems sufficient, and would be more consistent with other failures.

johnstevenson:
> Looks like ReadFile failure isn't handled correctly

In addition to the uninitialized bytes, simulating a ReadFile failure with some data read into the buffer (I don't know how realistic this acutally is though), results in a single stream of unparsed bytes in the buffer. `if FOkayToRead and LastRead ...` is one way to fix it.

johnstevenson:
> One thing I'm not quite understanding, and this was in the previous PR too, is why, when CreatePipe or SetHandleInformation fail, the exception isn't raised immediately, but is instead delayed until UpdateStartupInfo is called.

Ah, good point. I'll fix this by raising the exception in the constructor.

> Also why does it log the error and raise an exception? I believe in most cases that will result in the error appearing in the log twice. Just raising the exception seems sufficient, and would be more consistent with other failures.

It will only appear twice if the user catches and logs the exception. I was following the semantics of ExecAndLogOutput, but I will change this to make failures more consistent.

martijnlaan:
> > Looks like ReadFile failure isn't handled correctly
> 
> In addition to the uninitialized bytes, simulating a ReadFile failure with some data read into the buffer (I don't know how realistic this acutally is though), results in a single stream of unparsed bytes in the buffer. `if FOkayToRead and LastRead ...` is one way to fix it.

My suggestion would be to call `SetLength` again to truncate the buffer back to its size before the error so `SetLength(FReadBuffer, TotalBytesHave`. Like this:

```
if not FOKToRead then begin
        LogErrorFmt('ReadFile failed (%d).', [GetLastError]);
        SetLength(FReadBuffer, TotalBytesHave); { Restore back to original size }
```

johnstevenson:
The latest fafba1ec fixes recent suggestions.

I note that `TCreateProcessOutputReader.MaxTotalBytesToRead` is not used and perhaps should be removed?

> Yeah, 2MB does sound plenty adequate for typical console utilities. What I'm not sure about are atypical uses where someone might want to get a dump of data, as opposed to collecting status or error messages for debugging purposes.

I was actually basing the 2MB on a dump of data. Following on from [Jordan's example data](https://github.com/jrsoftware/issrc/pull/468#issuecomment-2199334508), I've produced a table showing the effect of line length:

| Output | Characters |     Lines | List/Array |
|--------|:----------:|----------:|-----------:|
| 1 mb   | 1          |   524,288 |   18.00 mb |
| 1 mb   | 16         |    61,681 |    3.88 mb |
| 1 mb   | 32         |    31,775 |    2.97 mb |
|        |            |           |            |
| 5 mb   | 1          | 2,621,440 |   90.00 mb |
| 5 mb   | 16         |   308,405 |   19.41 mb |
| 5 mb   | 32         |   158,875 |   14.85 mb |

Characters is the number of characters in a line of output (not including a LF separator), lines is the number of lines the output produces and List/Array is the memory needed to populate the string list and then the array. Of course this memory will increase when the array is being populated from the string list, to double the amount shown.

However, I still have no idea what the maximum output value should be! And also if it should be the same as or different to `ExecAndLogOutput`, where TOnLog could be used to write each line to a string list or array.

Projects/Src/CmnFunc2.pas
```diff
@@ -1597,14 +1611,37 @@ function TOneShotTimer.TimeRemaining: Cardinal;
...
+    FRedirectionError := Format(S, Args);
+  end;
+
+  function PipeCreate(var Read, Write: THandle; SecurityAttr: TSecurityAttributes): Boolean;
+  begin
+    Result := False;
+    if not CreatePipe(Read, Write, @SecurityAttr, 0) then
+      SetRedirectionError('CreatePipe failed (%d).', [GetLastError])
+    else if not SetHandleInformation(Read, HANDLE_FLAG_INHERIT, 0) then
+      SetRedirectionError('SetHandleInformation failed (%d).', [GetLastError])
```
jordanrussell, line 1627: Sorry, what I meant was: why doesn't it raise the exceptions at the moment the errors occur? That way you wouldn't need the `FRedirectionError` string. And, because an exception stops further execution of code in the constructor, the `else` and the Boolean function result would no longer be necessary.

So something like this?
```pascal
  procedure PipeCreate(var Read, Write: THandle; const SecurityAttr: TSecurityAttributes);
  begin
    if not CreatePipe(Read, Write, @SecurityAttr, 0) then
      raise Exception.CreateFmt('Output redirection error: CreatePipe failed (%d).', [GetLastError]);
    if not SetHandleInformation(Read, HANDLE_FLAG_INHERIT, 0) then
      raise Exception.CreateFmt('Output redirection error: SetHandleInformation failed (%d).', [GetLastError]);
  end;
```

Projects/Src/CmnFunc2.pas
```diff
@@ -1706,44 +1788,57 @@ procedure TCreateProcessOutputReader.Read(const LastRead: Boolean);
...
-        LogErrorFmt('ReadFile failed (%d).', [GetLastError])
-      else if BytesRead > 0 then begin
+      if not FOKToRead then begin
+        LogErrorFmt('ReadFile failed (%d).', [GetLastError]);
+        { Restore back to original size }
+        SetLength(Buffer, TotalBytesHave);
+      end else if BytesRead > 0 then begin
         { Correct length if less bytes were read than requested }
-        SetLength(FReadBuffer, TotalBytesHave+BytesRead);
+        SetLength(Buffer, TotalBytesHave+BytesRead);
```
jordanrussell, line 1802: To complete the fix, I believe `else if BytesRead > 0 then` needs to be removed.

Currently, if `ReadFile` returns `True` and `BytesRead = 0`, both of the above `SetLength` calls are bypassed.

I can't think of a scenario in which `BytesRead` would be `0` when `PeekNamedPipe` said there were bytes available. But better safe than sorry.

jordanrussell:
> I note that `TCreateProcessOutputReader.MaxTotalBytesToRead` is not used and perhaps should be removed?

I guess Martijn added that to allow the maximum size to be adjusted externally after the instance was created?

(But you wouldn't want to lower it after reading has begun because that could trigger an integer underflow in `DoRead`...)

> Of course this memory will increase when the array is being populated from the string list, to double the amount shown.

Unless ROPS is doing something weird, assigning the strings from the string list to the array should simply increment the reference counts from 1 to 2, so no new strings should be allocated.

> However, I still have no idea what the maximum output value should be!

Here's an idea: what if there were separate limits for bytes and lines? Like maybe 10M bytes but only 1M lines. That would allow large amounts of "legitimate" data to be read, while protecting against excessive memory/CPU usage in the 1-character-per-line case.

Projects/Src/CmnFunc2.pas
jordanrussell: Oh, another issue: If `CreatePipe` succeeds but `SetHandleInformation` fails, the two pipe handles will be leaked.

Looks like `PipeCreate` should be passed `FStdOutPipeRead` and `FStdOutPipeWrite` directly in the first two parameters. But, if that is done, `PipeCreate` must be careful not to modify the parameters unless `CreatePipe` succeeds (because the docs seem to suggest that garbage could be written there if it fails). By that I mean something like:

```pascal
  var TempReadPipe, TempWritePipe: THandle;
  if not CreatePipe(TempReadPipe, TempWritePipe, ...) then
    raise ...
  Read := TempReadPipe;
  Write := TempWritePipe;
```

martijnlaan:
I didn't have a particular reason in mind when I added MaxTotalBytesToRead as a property, removing is fine with me.

johnstevenson:
The latest a2f8d6bfcaf8939febfb527f105e44abecfdf292 retricts the output to 10 million bytes or 1 million lines, plus other fixes.

> Unless ROPS is doing something weird, assigning the strings from the string list to the array should simply increment the reference counts from 1 to 2, so no new strings should be allocated.

Indeed, I am a muppet for not reading the code to check.

martijnlaan:
Do you have a suggestion for the whatsnew entry John? Feel free to put it in there, also for the previous changes.

jordanrussell:
I went ahead and fixed a couple of things, and decided to make it non-fatal once again if NUL cannot be opened (based on your comments in #467).

I *think* that covers all the issues. Only remaining thing I'm not sure about is whether throughput could be improved by adjusting the pipe buffer size or sleep time (as mentioned above). But that's really independent of this PR.

martijnlaan:
> I went ahead and fixed a couple of things

I'm suprised you were able to commit on John's branch. Was that allowed by default?

jordanrussell:
I don't know if it's on by default, but there is an option to give upstream permission to commit:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

martijnlaan:
> I don't know if it's on by default, but there is an option to give upstream permission to commit: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

Interesting, thanks.

johnstevenson:
> I don't know if it's on by default, but there is an option to give upstream permission to commit

When you create a pull request, there is an _Allow edits from maintainers_ option:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork

I usually select this because it makes things easier for maintainers.

Do you want me to rebase and squash this (leaving Jordan's three commits intact) , or will you do it?

martijnlaan:
I will merge this but without any squashes because I prefer to keep the history as much as possible.

martijnlaan (approved):

johnstevenson:
> ... and decided to make it non-fatal once again if NUL cannot be opened

But this means that an error will be set in `TOnLog` and `TExecOutput`, which would indicate  a failure condition to the user, even though it is expected to succeed?

```
Error will be True if reading the output failed (which should be very rare), or if the output is too big.
There is no further output after an error.
```

jordanrussell:
Oops, that's right. I've removed the logging.

I looked at possibly extending `TLogProc` to accept "warnings" as well, which would be written to Setup's log but not passed to the user's log procedure. But it didn't seem worth it, since it would only be used in this one case, and for an error condition that shouldn't actually cause a problem.

martijnlaan:
I've merged this. Thanks for the effort!

Did a bit of minor cleanup and fixed the example in 3dc589000e5216df1689d4599a30b478c4cdaf33 and 65b9a9d379cea2c7a422e67968aeada83e38e513.

PR #472, opened 2024-07-06
Clean the pipe code, also fixing the issues.

martijnlaan:


johnstevenson:
The addition of `TCreateProcessOutputReaderData` makes it so much cleaner.

I forgot to bring this up in the original PR, but I was wondering whether `ERR_BROKEN_PIPE` should be checked for after a ReadFile failure?

martijnlaan:
> The addition of `TCreateProcessOutputReaderData` makes it so much cleaner.
> 
> I forgot to bring this up in the original PR, but I was wondering whether `ERR_BROKEN_PIPE` should be checked for after a ReadFile failure?

To skip the HandleAndLogErrorFmt?

johnstevenson:
Yes, to skip logging it as an error.

jordanrussell:
I changed it to disregard any incomplete line when `PeekNamedPipe` or `ReadFile` fail because the help says there isn't supposed to be further output after an error (and that makes sense to me). And I added proper handling of `ERROR_BROKEN_PIPE` from `ReadFile` in the process (even though that should be impossible since `PeekNamedPipe` said there were bytes waiting).

jordanrussell:
One more issue came to mind, and I think it's in the 6.3 code as well:

- Child process writes only 2 characters to stdout: `'A'#13`
- `Read(False)` reads those 2 characters into `Buffer`, but doesn't process the line since the buffer ends with `#13`.
- `Read(True)` finds the pipe is empty, so the `Check for completed lines` code is skipped. At the end of the function, it calls `LogLine(Buffer)` -- but the `#13` is still in the buffer. Oops.

I haven't tested this to confirm, so there may be something I'm missing.

If not, then:
- should it check for and strip off a `#13` before calling `LogLine`? (simplest fix?)
- should the `Check for completed lines` code be moved? (quite invasive?)

jordanrussell:
btw, I could not stop laughing at the search & replace gone awry:
```
{ Append newly available Pipe to the incomplete line we might already have }
{ Check for completed lines thanks to the new Pipe }
```

martijnlaan:
>     * should it check for and strip off a `#13` before calling `LogLine`? (simplest fix?)

I made it strip the #13 but then also log a second empty line.

johnstevenson:
> I made it strip the #13 but then also log a second empty line.

Adding a second empty lines feels wrong to me because it doesn't match the `FindNewLine` comment:

```
This will return the position of the first #13 or #10. If a #13 is at
the very end of the string it's only accepted if we are certain we can't
be looking at a split #13#10 because there will be no more reads
```

So in the normal flow (if Pipe.OkToRead and LastRead are True), then a single line is logged with the #13 is removed:

```delphi
LogLine(Pipe.CaptureList, Copy(Pipe.Buffer, 1, P-1));
```

martijnlaan:
Ah, yes, thanks.

PR #475, opened 2024-10-30
Add specific environment variable for temp folder

michelcedric:
Add Support of a new environment variable to have to possibility to use specific folder for temp file of inno setup

### DIFF
--- a/Projects/Src/Shared.CommonFunc.pas
+++ b/Projects/Src/Shared.CommonFunc.pas
@@ -754,6 +754,9 @@ function GetTempDir: String;
   differences. }
 label 1;
 begin
+  Result := GetEnv('INNOSETUPTMP');
+  if (Result <> '') and DirExists(Result) then
+    goto 1;
   Result := GetEnv('TMP');
   if (Result <> '') and DirExists(Result) then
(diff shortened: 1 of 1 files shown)


martijnlaan:
Thanks but GetTempDir returns the system's temp dir and not Inno Setup's temp dir. And even if it did I don't understand why this would be a good and safe change,

michelcedric:
@martijnlaan 
I want to change the temp dir use by Inno setup.
I want to propose to have the possibility to add an another specific folder by the usage of a specific environment variable.

martijnlaan:
Well, you definitely shouldn't since the dirs it currently uses are the safest choice, except perhaps for placing them inside the system temp dir itself when possible. Allowing some kind of other dir seems unsafe.

jordanrussell:
> We want to avoid to change TMP value because other application use it.

You don't have to change the value globally. Processes inherit environment variables from their parent, so you could write a simple batch file that sets the variable and then runs the setup EXE:

```bat
set TMP=%LOCALAPPDATA%\SpecialTempDir
mkdir "%TMP%"
mysetup.exe
```

PR #477, opened 2024-11-18
Update Lithuanian translation for 6.4.0+

Tichij:


martijnlaan:
Thanks! I've merged your update to a branch. Merging it to main would automatically update the website as well which is a bit too early.

PR #487, opened 2024-12-12
Updated Swedish translation

yeager:
Reviewed and updated Swedish translation. Please use as Official and I will update it when needed,


martijnlaan:
Thanks!

PR #498, opened 2025-01-27
Azerbaijan (6.4.0).islu

miryusifrahimov:
6.4.0


martijnlaan:
Thanks! Please note the extension should be .isl so I have changed that. I have closed the other 2 PR's without using the files because only the latest version of a translation is made available.

miryusifrahimov:
I wrote the names like that myself, to avoid confusion, I said you can change them.

PR #513, opened 2025-05-21
Fix translation file encodings / LanguageCodePage

past-due:
This PR does 3 things:
1. Fixes `LanguageCodePage=0` for files that are already UTF8 with BOM
2. Fixes `LanguageCodePage=0` for files that are already UTF8 (without BOM) and prepends the BOM
3. Converts ANSI language files to UTF8 (w/ BOM) and fixes `LanguageCodePage=0`

After this PR, every official language translation file will be:
- UTF8-encoded, with `LanguageCodePage=0`, and UTF-8 BOM

### Notes:
- Why UTF8 with BOM?
  1. UTF8 files without a BOM "are not supported before Inno Setup 6.3"
  2. GitHub's web editor doesn't convert ANSI files as is desired (ref: [CONTRIBUTING.md](https://github.com/jrsoftware/issrc/blob/45053d7f163cb9d3f75317b1c5717ebd38dff4cc/CONTRIBUTING.md#contributing-translations))
  3. Helps ensures that InnoSetup *and other tools and APIs that edit INI files* can reliably detect the encoding of the files.
     - This even includes the `Write­Private­Profile­StringW` (etc) Windows APIs, which will properly respect the existing Unicode-ness of a file when writing to it, as long as they can reliably detect it's Unicode. [The BOM serves as a big hint.](http://archives.miloush.net/michkap/archive/2006/09/15/754992.html#756283)
- Why convert all the translations?
    - Standardizing on UTF-8 improves the editing and viewing experience for the files
    - Makes the process of working with layered custom `.isl` files easier (Example: In the case of using multiple `MessagesFile`(s) in a `[Languages]` section directive, when you want to use the base translations and also have a separate set of translated custom messages. Most everything outputs UTF-8 encoded files these days, and if the base (first) translation .isl is using ANSI encoding...)

### Reviewer Notes:

GitHub's web display of diffs seems to handle (most?) ANSI codepage detection, so it ends up showing changed lines that appear to be the same. The underlying encoding of the files is different (ANSI codepage -> UTF8). Although I suppose this helps visually checking that the conversions resulted in the same output.

### Conversion Script:

Below is the custom PowerShell script used to do the conversions.

<details>
<summary>Click to view PowerShell script</summary>

```powershell
function Get-IniValue {
    param(
        [string]$FilePath,
        [string]$Section,
        [string]$Key
    )

    try {
        $content = Get-Content -Path $FilePath -ErrorAction Stop
    }
    catch {
        Write-Error "Error reading file: $($_.Exception.Message)"
        return $null
    }

    $sectionFound = $false
    foreach ($line in $content) {
        if ($line -match '^\s*\[(.*)\]\s*$') {
            $currentSection = $matches[1]
            if ($currentSection -eq $Section) {
                $sectionFound = $true
            }
            else {
                $sectionFound = $false
            }
        }
        elseif ($sectionFound -and $line -match '^\s*(.*?)\s*=\s*(.*?)\s*$') {
            $currentKey = $matches[1]
            $currentValue = $matches[2]
            if ($currentKey -eq $Key) {
                return $currentValue
            }
        }
    }

    # Key not found in section, or section doesn't exist
    return $null
}

function Is-ValidUTF8 {
    param(
        [byte[]]$bytes
    )
    try {
        [System.Text.UTF8Encoding]::new($false, $true).GetString($bytes) | Out-Null
        return $true
    }
    catch {
        return $false
    }
}

function Convert-AnsiLanguageFileToUtf8Impl {
    param(
        [Parameter(Mandatory=$true)]
        [string]$InputFilePath,
        [Parameter(Mandatory=$true)]
        [string]$OutputFilePath
    )

    # Get current LanguageCodePage from .isl file LangOptions section
    $inputCodePageInt = 0
    $languageCodePageStrVal = Get-IniValue "$InputFilePath" "LangOptions" "LanguageCodePage"
    if ([string]::IsNullOrEmpty($languageCodePageStrVal)) {
        Write-Host "`tInput file: '$InputFilePath' missing LanguageCodePage=, defaulting to LanguageCodePage=0"
    }
    elseif ($languageCodePageStrVal -eq '0') {
        Write-Host "`tInput file: '$InputFilePath' already has UTF-8 LanguageCodePage=0"
        return # skip any processing
    }
    else {
        Write-Host "`t'$InputFilePath': LanguageCodePage=$inputCodePageInt"
        $inputCodePageInt = [System.Convert]::ToInt32($languageCodePageStrVal)
    }

    $convertedBytes;
    $inputBytes = Get-Content "$InputFilePath" -Raw -Encoding Byte
    if ($inputBytes.Length -ge 3 -and $inputBytes[0] -eq 0xEF -and $inputBytes[1] -eq 0xBB -and $inputBytes[2] -eq 0xBF) {
        # File already has the UTF-8 BOM, just fix-up the LanguageCodePage
        Write-Host "`t'$InputFilePath': already has UTF-8 BOM - treating as UTF8 and fixing up LanguageCodePage"
        $convertedBytes = $inputBytes[3..($inputBytes.Length - 1)]
    }
    elseif (Is-ValidUTF8 $inputBytes) {
        Write-Host "`t'$InputFilePath': already appears to be valid UTF-8 - treating as UTF8 and fixing up LanguageCodePage"
        $convertedBytes = $inputBytes
    }
    else {
        Write-Host "`t'$InputFilePath': converting from ANSI codepage $inputCodePageInt to UTF-8"
        $ansiEncoding = [System.Text.Encoding]::GetEncoding($inputCodePageInt)
        $convertedBytes = [System.Text.Encoding]::Convert($ansiEncoding, [System.Text.UTF8Encoding]::new($false, $true), $inputBytes)
    }
    $outputString = [System.Text.Encoding]::UTF8.GetString($convertedBytes);

    if (-not [string]::IsNullOrEmpty($languageCodePageStrVal)) {
        $lineToReplace = "LanguageCodePage=$languageCodePageStrVal"
        $replacementLine = "LanguageCodePage=0"
        $outputString = $outputString -replace [regex]::Escape($lineToReplace), $replacementLine
    }
    else {
        Write-Warning "$OutputFilePath will lack explicit LanguageCodePage=0 line, because input file lacked one"
    }

    Write-Host "`tWriting '$OutputFilePath' as (UTF8 with BOM), LanguageCodePage=0"
    $utf8EncodingWithBom = [System.Text.UTF8Encoding]::new($true, $true)
    [System.IO.File]::WriteAllText($OutputFilePath, $outputString, $utf8EncodingWithBom)
}

function Convert-AnsiLanguageFileToUtf8 {
    param(
        [Parameter(Mandatory=$true, ValueFromPipeline=$true)]
        [System.IO.FileSystemInfo]$Item
    )
    process {
        Write-Host "Processing: $($Item.FullName)"
        Convert-AnsiLanguageFileToUtf8Impl "$($Item.FullName)" "$($Item.FullName)"
    }
}

Get-ChildItem ".\*.isl" | Convert-AnsiLanguageFileToUtf8
````

</details>


martijnlaan:
Thanks but there are multiple problems with this PR. BOMs are not desirable. Many if not most maintainerz don't use Git and these changes would get last. Setting LanguageCode to 0 isn't fixing anything.

past-due:
> Thanks but there are multiple problems with this PR. BOMs are not desirable. Many if not most maintainers don't use Git and these changes would get lost. Setting LanguageCode to 0 isn't fixing anything.

To clarify: The changes shouldn't get lost as almost all tools - including text editors, code editors, and anything using the Windows APIs to edit INI files - will properly detect and maintain the file encoding (for sure if the files have the BOM, and still very likely - but not 100% certain - without the BOM)?

(The LanguageCode=0 change was only made as a result of fixing the actual encoding of the files to UTF-8.)

Since this PR addresses several inconsistencies between what the files say (LanguageCodePage) and their actual encoding, as well as attempts to standardize the coding to make editing easier, are there portions that would you be willing to accept?

Specifically, the current situation is that:
1. Some `.isl` files are _already actually UTF-8 encoded_ (with or without a BOM) **but** specify a non-0 `LanguageCodePage` - surely it would make sense to correct the `LanguageCodePage` to 0 on these?
2. Is there any way you'd accept converting the ANSI `.isl` files to Unicode? (UTF8 with BOM was selected because the `CONTRIBUTING.md` explicitly calls this out: "`UTF8 files without a BOM [...] are not supported before Inno Setup 6.3`". And also because it ensures that other tools properly handle it - even if they directly use the Windows APIs - so that translators would have no issues.)

martijnlaan:
- I meant the changes would likely get lost when a maintainer who doesn't use Git would submit an update to their translation via the submit form on the website. To get a translation updated you would have to contact them via email instead of using a PR, even if it's a 6.4.0+ translation.
- Changing a language file's LanguageCodePage to 0 breaks [Languages] entries for that language which use non Unicode license and info files, so I don't think this change is ever OK.
- The CONTRIBUTING.md note does not mean to say files with a BOM are preferred in Inno Setup because they are not, instead files without one are. It also does not mean to say updated translations are expected to somehow support old versions because they are not and in fact can't because the compiler of the old version will raise a fatal error on encountering new messages.

Doing a PR to remove BOMs for all 6.4.0+ .isl files for which the Git history shows they have been using PRs themselves would be OK. I don't have a list myself who does and doesn't though. Same for changing them non Unicode to UTF-8.

PR #535, opened 2025-06-23
Polish.isl - translation update for 6.5.0

Hirotaro:
Translation update following the request with some minor tweaks


martijnlaan:
Thanks! Am I seeing it right that the updates for DownloadingLabel -> DownloadingLabel2 and ExtractionLabel -> Extracti**ng**Label2 are missing?

Hirotaro:
@martijnlaan you are correct, I have missed that, was looking only on + and - lines. 
I have fixed those 2 lines in latest commit.

PR #546, opened 2025-07-02
Update French.isl

zedalaye:
Fixed verb tuning issue


zedalaye:
Thatks to [github.com/jogo_](https://github.com/jogo-)

PR #547, opened 2025-07-04
Update Catalan.isl

Ecron:
Updated Catalan translation to latest update (by Artur aka Ecron at GitHub).


martijnlaan:
Thanks! I did two small tweaks, see 95e9d72e

PR #549, opened 2025-07-12
Update Russian translation for 6.5.0+

yktoo:


yktoo:
Thanks @martijnlaan, I've reset my fork's `main` to your upstream, so I'm perfectly aligned with the upstream as of now. UTF-8 is indeed way better going forward.

PR #559, opened 2025-09-28
Italian language improvements

bovirus:
@martijnlaan 

Please merge. Thanks.

### DIFF
--- a/Files/Languages/Italian.isl
+++ b/Files/Languages/Italian.isl
@@ -9,7 +9,7 @@
 ; Setup adds the periods automatically (appending a period would result in
 ; two periods being displayed).
 ;
-; Italian.isl - Last Update: 22.06.2025  by bovirus (bovirus@gmail.com)
+; Italian.isl - Last Update: 28.09.2025  by bovirus (bovirus@gmail.com)
 ;
 ; Translator name:   bovirus
 ; Translator e-mail: bovirus@gmail.com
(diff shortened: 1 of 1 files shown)


bovirus:
@martijnlaan 

Why did you close PR?
I can update the file with your change request and you can appove this.
I update the language with you reuest.
Pelase re-open PR and apply change.

I believe that you can consider that in the spèecfic context "version" could be an info that didn't add significative info.

With the original language the caption of main window of the installer is like

- Installer of Inno Setup version 6.6.0

With my change the main window caption is like

- Installer of Inno Setup 6.6.0

I believe that "version" in this context didn't add an imporat adding to the string.

martijnlaan:
I closed it because I couldn't merge it and assumed you could reopen it once you did the update.

Please see translations page at https://jrsoftware.org/files/istrans/send.php, it's very clear:

'When translating the text, do not add to or change the meaning of messages to suit your personal taste. [...] All messages in your translation must be as close to the original English messages as possible.'

So you need to add 'version' back please. If there are more strings which have this problem, please also update those.

bovirus:
@martijnlaan 

I will modify the language as you requested.

For me remove "version" didn't chnage the meaning and didn't remove something to meaning of the string.
Version is only an additional specfication not necessary in this context because is clear that number after program name is the version.
And could be true for all languages.

I will re-open a new PR.

PR #560, opened 2025-10-05
Add Vietnamese language translation

memecoder12345678:


martijnlaan:
Thanks! I've commited this as a replacement for the Vietnamese translation which already existed in the Unofficial dir, and have also assumed you used this as a base. If you mean the file to become an official translation, please see https://jrsoftware.org/files/istrans/send.php

I've also made three changes to the file myself. Please make sure you update your local copy:

- Changed line endings from Unix back to Windows
- Restored the 42A language id as 429 seems incorrect
- Restored two deleted comments at the top of the file

PR #592, opened 2026-04-28
Avoid creating empty registry key when CLI is reading SignTools

albertony:
Currently, running ISCC.exe creates an empty registry key `HKEY_CURRENT_USER\Software\Jordan Russell\Inno Setup` - if not already exists. I think this is unnecessary - in a stand-alone, portable install, where only the CLI is being used (e.g. in a CI workflow). This PR avoids it. The result makes the CLI more closer to what some refer to as a "stealth" or "portable app" tool.

### DIFF
--- a/Projects/Src/Shared.ConfigIniFile.pas
+++ b/Projects/Src/Shared.ConfigIniFile.pas
@@ -19,24 +19,40 @@ type
   private
     FMutex: THandle;
     FAcquiredMutex: Boolean;
+    procedure AcquireMutex;
   public
     constructor Create;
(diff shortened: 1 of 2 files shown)


martijnlaan:
Thanks. The PR doesn't work for me because of two issues:

- KEY_READ does not stop TRegIniFile from wanting to create the key
- ISCC doesn't only read sign tools but also license state.

I added a workaround for the first issue, and fixed the second.

Thanks for the suggestion 👍

albertony:
Sorry for that, and thank you very much!

PR #595, opened 2026-05-19
Update Italian.isl

bovirus:
@martijnlaan 

Please merge. Thanks.

### DIFF
--- a/Files/Languages/Italian.isl
+++ b/Files/Languages/Italian.isl
@@ -9,7 +9,7 @@
 ; Setup adds the periods automatically (appending a period would result in
 ; two periods being displayed).
 ;
-; Italian.isl - Last Update: 02.05.2026  by bovirus (bovirus@gmail.com)
+; Italian.isl - Last Update: 18.05.2026  by bovirus (bovirus@gmail.com)
 ;
 ; Translator name:   bovirus
 ; Translator e-mail: bovirus@gmail.com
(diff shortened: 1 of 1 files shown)


martijnlaan:
Thanks for the update. However, some changes go against the guidelines (https://jrsoftware.org/files/istrans/send.php): 

> When translating the text, do **not** add to or change the meaning of messages to suit your personal taste.

The NameAndVersion change in particular cannot be merged, but others like the AboutSetupMessage change also have this issue.

bovirus:
@martijnlaan

I beleieve that I didn't change the meaning of the string.

The word "versione" is not useful in this context because didn't add nothing that require the presence of "versione".

In software environment is not usual separate AppName and AppVersion with strings version

Then "vesrione" is a word that is not necessary. I didn't see any additional benefit in

Installazione di Inno Setup **versione** 6.7.2 (current)

then 

Installazione di Inno Setup 6.7.2 (my proposal - more modern and without string not necessary.

Of course I talk about inno Setup 6.xx.

martijnlaan:
Yes, I understand what you're saying and it's a good idea. I will look into updating Default.isl to not use 'X version Y' at some point.

For now it still does, though, which means translations should follow the same structure. Wherever Default.isl says 'version', Italian.isl should also have that word translated, not removed.

bovirus:
@martijnlaan 

Thanks for the oppurtinity to consider this change for all languages (I hope soon...).

PR #597, opened 2026-06-19
TNewCheckListBox upgrade

leserg73:
Updating the TNewCheckListBox component.
P.S. Sorry, Google translation.

New _properties_:
 - **ItemFontColor** - set font color for Item
 - **SubItemFontColor** - set font color for subitem
 - **UseStyledColor** - On/Off custom color if VCL Style use. Default set is False
 - **TreeViewStyle** - displaying the list as a tree view
 - **ShowRoot** - displaying root nodes
 
New _methods_:
- **CollapseAll** - collapses all nodes
- **ExpandAll** - expands all nodes
- **CollapseItem** - collapse desire nodes
- **ExpandItem** - expand desire nodes
- **AddCheckBoxEx**, **AddGroupEx**, **AddRadioButtonEx** - adding an element that contains child elements as a collapsed element

New _events_:
- **OnExpandCollapse** - occurs when a node is expanded or collapsed
- **OnItemMouseMove** - occurs when the mouse cursor hovers over a list item
... and added disabled events:
- OnMouseDown, OnMouseEnter, OnMouseLeave, OnMouseMove, OnMouseUp.

New _directive_:
- **TreeComponentsList** - allows to display a list of components in a tree view mode with +/- buttons

New _flag_ (for Components section):
 - **collapsed** - allows to specify that a collapsed node should be created if TreeComponentsList=yes

You can test the updated TNewCheckListBox component using the following Inno Setup scripts:
Components_ColoredCaption.iss
Components_OnItemMouseMove.iss
Components_TVStyle.iss
Look it in archive file:
[TNewCheckListBox_up.zip](https://github.com/user-attachments/files/29150105/TNewCheckListBox_up.zip).

### DIFF
--- a/Components/NewCheckListBox.pas
+++ b/Components/NewCheckListBox.pas
@@ -26,6 +26,7 @@ const
 type
   TItemType = (itGroup, itCheck, itRadio);
   TCheckBoxState2 = (cb2Normal, cb2Hot, cb2Pressed, cb2Disabled);
+  TExpandButtonState = (ebsCollapsed, ebsExpanded);
 
   TItemState = class(TObject)
   public
@@ -44,6 +45,8 @@ type
     ItemFontStyle: TFontStyles;
     SubItemFontColor: TColor;
     SubItemFontStyle: TFontStyles;
+    Expanded: Boolean;
+    HasChildren: Boolean;
   end;
 
   TCheckItemOperation = (coUncheck, coCheck, coCheckWithChildren); 
@@ -72,9 +75,16 @@ type
     FDisableItemStateDeletion: Integer;
     FDisableStyledButtons: Boolean;
     FUseStyledColor: Boolean;
+    FTreeViewStyle: Boolean;
+    FExpandButtonSize: Integer;
+    FExpandButtonColor: TColor;
+    FExpandButtonLineColor: TColor;
+    FLastMouseOverIndex: Integer;
+    FHasAnyChildren: Boolean;
     class constructor Create;
     class destructor Destroy;
     class var FComplexParentBackground: Boolean;
+    procedure UpdateHasAnyChildren;
     procedure UpdateThemeData(const Close, Open: Boolean);
     function CanFocusItem(Item: Integer): Boolean;
     function CheckPotentialRadioParents(Index, ALevel: Integer): Boolean;
@@ -88,7 +98,7 @@ type
     procedure EndCapture(Cancel: Boolean);
     function AddItem2(AType: TItemType; const ACaption, ASubItem: string;
       ALevel: Byte; AChecked, AEnabled, AHasInternalChildren,
-      ACheckWhenParentChecked: Boolean; AObject: TObject): Integer;
+      ACheckWhenParentChecked, AExpanded: Boolean; AObject: TObject): Integer;
     function FindAccel(VK: Word): Integer;
     function FindCheckedSibling(const AIndex: Integer): Integer;
     function FindNextItem(StartFrom: Integer; GoForward,
@@ -112,6 +122,11 @@ type
     procedure WMThemeChanged(var Message: TMessage); message WM_THEMECHANGED;
     procedure WMUpdateUIState(var Message: TMessage); message WM_UPDATEUISTATE;
     procedure WMVScroll(var Message: TWMVScroll); message WM_VSCROLL;
+    function GetExpandButtonRect(Index: Integer): TRect;
+    procedure ToggleExpand(Index: Integer);
+    function HasVisibleChildren(Index: Integer): Boolean;
+    procedure SetTreeViewStyle(Value: Boolean);
+    function ShouldShowExpandButton(Index: Integer): Boolean;
   protected
     procedure CreateWnd; override;
     procedure MeasureItem(Index: Integer; var Height: Integer); override;
@@ -221,6 +236,7 @@ type
     property UseStyledColor: Boolean read FUseStyledColor write SetUseStyledColor default False;
     property Visible;
     property WantTabs: Boolean read FWantTabs write FWantTabs default False;
+    property TreeViewStyle: Boolean read FTreeViewStyle write SetTreeViewStyle default False;
   end;
 
   TNewCheckListBoxStyleHook = class(TScrollingStyleHook)
@@ -244,7 +260,7 @@ procedure Register;
 implementation
 
 uses
-  UITypes, Types, ActiveX,
+  UITypes, Types, ActiveX, Math,
   NewUxTheme.TmSchema, PathFunc, BidiUtils, UnsignedFunc;
 
 const
@@ -439,6 +455,7 @@ begin
     end;
   end;
 
+  FHasAnyChildren := False;
   FStateList := TList.Create;
   FMinItemHeight := 16;
   FOffset := 4;
@@ -446,7 +463,13 @@ begin
   Style := lbOwnerDrawVariable;
   FHotIndex := -1;
   FCaptureIndex := -1;
+  FLastMouseMoveIndex := -1;
   FUseStyledColor := False;
+  FTreeViewStyle := False;
+  FExpandButtonSize := 9;
+  FExpandButtonColor := clBtnFace;
+  FExpandButtonLineColor := clWindowText;
+  FLastMouseOverIndex := -1;
 end;
 
 procedure TNewCheckListBox.CreateWnd;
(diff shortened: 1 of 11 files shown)


martijnlaan:
Thanks for your work 👍 Wish you mentioned it to us sooner, so feedback could be offered sooner as well. Are you still interested in review findings and updating the PR for this?

Here's some thoughts I'm having now, which includes a showstopper:

- Does it work with right to left languages, with high DPI and with VCL styles, including excludelightcontrols?
- Some commits seem unrelated to having a collapsable list. These should be extracted. Some things I notice
  - Unrelated fixes like 0da22222096fa05cf358428209e8c58f1ab94ef3 and  fd0219e45735580dc0e3467b4ffcb0967f305d0f. I will likely cherry pick those myself into `vnext` already. Edit: actually I think fd0219e is against how VCL Styles work? ParentColor is not supposed to have an effect like that?
  - Unrelated additions of font color properties in 881c49b107e8ef49c803ebfd0fe663d621ea030f
  - Unrelated addition of UseStyledColor property. That's not how VCL Styles work and unlike all other components, also in 881c49b107e8ef49c803ebfd0fe663d621ea030f.
  - Unrelated addition of MouseOver code in 662cb8b860952dc54c4dcee261f8e16401b48967.
  - 7edbeac4c34ab7e360b03f4bb06da1c8f44accd3 is also not how VCL Styles are supposed to work and is unlike all other components, and breaks runtime styling changes? Not sure if it's related or if the collapse code needs this.
  - Same for 67c6ae6fbe025e36af9d7f2b0653489e20ae5a79: unlike other components. What was the result of the further testing it mentions?
- The PR isnt based on `vnext` branch, but it should be.
- Adding so many `Ex` functions should probably avoided. The caller can use the result index instead?
- The initial 3 commits were signed but the rest was not, what happened and can you reenable signing?
- How does it work together with `WantTabs`?
- Can you offer a description on how the collapsing works in your own words? Instead of me having to deduce this from th ecode. From the review findings below it is my impression collapsing works by making an item's height 1, but that means you could still interact with it? And multiple collapsed items in a row would lead to bigger gaps? If a collapsed node hides
  200 descendants, there is a huge 200 pixel gap? Or am I misunderstanding. Seems like a showstopper.
- Can you offer a description of the use case? I suppose it's needed only when there's a huge amount of components? Does previous note not contradict with that? Have you tested performance with huge lists?

Also some review findings from Claude and Codex for f21f8d5ef79306b9a9fbc116dbd46f73411c7cf1:

---

| #   | Severity | Area        | One-line                                                         |
| --- | -------- | ----------- | ---------------------------------------------------------------- |
| 1   | Medium   | Performance | Per-item `RemeasureItem` + full-window invalidate on every add   |
| 2   | Medium   | Correctness | Enabling `TreeViewStyle` does not remeasure existing items       |
| 3   | Medium   | Correctness | Collapsed items are only shrunk to height 1, still hit-testable  |
| 4   | Medium   | Correctness | Space can toggle a hidden selected item after collapse           |
| 5   | Medium   | API misuse  | `LB_SETCOUNT` sent to a non-`LBS_NODATA` list box                |
| 6   | Low      | Correctness | Inconsistent `FHasAnyChildren` gating of the indent offset       |
| 7   | Low      | RTL         | Themed checkbox-size fallback inconsistent with new rect logic   |
| 8   | Low      | Clarity     | `ShouldShowExpandButton` ignores its `Index` parameter           |
| 9   | Low      | Correctness | `HasChildren`/`FHasAnyChildren` go stale when last child deleted |
| 10  | Low      | Dead code   | Unused field `FLastMouseOverIndex`                               |
| 11  | Style    | Conventions | `var` blocks, `begin` placement, comment style                   |

### Medium

**1. Per-item remeasure and full invalidate in `AddItem2`.**
`AddItem2` now calls `RemeasureItem` and `UpdateScrollRange` on every add (`1551`-`1554`), and `UpdateScrollRange` issues a full-client `InvalidateRect(Handle, nil, True)` (`735`). Populating N items becomes O(N) full invalidates plus repeated scroll-range churn, which can cause flicker and slowdown on large component lists. Consider batching, or guarding the per-item work so it only runs when not bulk-loading.

**2. Enabling `TreeViewStyle` does not remeasure existing items.**
`SetTreeViewStyle` (`2497`) only remeasures in the disable branch. `MeasureItem` (`776`) reserves the expand-button column whenever tree style is active, so items added before `TreeViewStyle := True` keep their old measurement: captions measured at the wider width are then drawn in the narrower width, clipping long or word-wrapped captions. The enable branch should also remeasure all items and update the scroll range.

**3. Collapsed items are only shrunk to height 1, not removed.**
`MeasureItem` returns height 1 for collapsed descendants (`756`). They still occupy a pixel each (inflating the scroll range and leaving gaps) and remain hit-testable. `MouseDown` (`1755`) uses `ItemAtPos` + `CanFocusItem` with no visibility check, so clicking the 1px sliver of a hidden item can still `Toggle` it. The visibility predicate added for navigation was not applied to the input paths.

**4. Space can toggle a hidden selected item after a collapse.**
Repro:
1. `TreeViewStyle` on, a parent with a checkable child.
2. Click the child, so `ItemIndex` = child.
3. Click the parent's `[-]` to collapse. `ToggleExpand` (`838`) never updates `ItemIndex`, so it still points at the now-hidden child.
4. Press Space. `KeyDown` (`1722`) gates only on `CanFocusItem(ItemIndex)`, and `CanFocusItem` (`558`) checks only `Enabled` and `ItemType <> itGroup`, not visibility, so it silently toggles the hidden item.

Arrow keys are safe because they route through `FindNextItem`, whose nested `IsItemVisible` (`1574`) skips collapsed descendants. Only the Space path bypasses that check. There is also a visible side effect: after collapse the selection highlight sits on an invisible 1px row, so the page shows no apparent selection while `ItemIndex >= 0`. Smallest fix: in `ToggleExpand`, when collapsing, move `ItemIndex` to the collapsed parent (or nearest visible focusable item) if it falls inside the collapsed subtree. More robust: promote `IsItemVisible` to a method and consult it in `KeyDown` and `MouseDown`.

**5. `LB_SETCOUNT` sent to a non-`LBS_NODATA` list box.**
`UpdateScrollRange` sends `LB_SETCOUNT` (`733`), but the control is created as `lbOwnerDrawVariable`. Per the Win32 docs this is invalid for non-`LBS_NODATA` list boxes (returns `LB_ERR`, no effect); the actual repaint is done by the added `InvalidateRect`. The surrounding comment still describes only the old "ineffectual `LB_SETTOPINDEX`" trick and was not updated. Recommend removing the `LB_SETCOUNT` call and refreshing the comment.

### Low

**6. Inconsistent `FHasAnyChildren` gating of the indent offset.**
The expand-button indent offset is gated by `... and FHasAnyChildren` in `CNDrawItem` (`692`, `700`) but not in `MeasureItem` (`776`), `InvalidateCheck` (`1713`), or `WMMouseMove` (`2386`). When `TreeViewStyle` is on but no item has children (`FHasAnyChildren = False`, for example a flat list), the checkbox is drawn without the offset while measurement and hit-testing assume it, misaligning clicks and hot-tracking.

**7. Themed checkbox-size fallback is inconsistent with the new rect logic.**
The classic-theme fallback for a non-standard `GetThemePartSize` result (`1343`-`1346`) still uses the old `Rect.Left - (Size.cx + FOffset)` plus `FlipRect` approach. The refactor removed the top-of-`DrawItem` undo-flip and made the main path compute RTL explicitly without flipping, so `Rect` is now physical. As a result this fallback can place the checkbox on the wrong side in RTL when the theme returns a size different from `FCheckWidth`/`FCheckHeight`.

**8. `ShouldShowExpandButton` ignores its `Index` parameter.**
It returns `FTreeViewStyle and not FWantTabs` for all items (`2512`), so despite the name it makes no per-item decision. This is what makes the gating in finding 6 easy to get inconsistent. Either drop the parameter or fold the per-item condition in.

**9. `HasChildren`/`FHasAnyChildren` go stale on child deletion.**
`LBDeleteString` calls `UpdateHasAnyChildren` only when the deleted item itself had `HasChildren`. A parent's `HasChildren` flag is never cleared when its last child is removed, so both flags can become inaccurate. Mostly masked today by `HasVisibleChildren` scanning live items, but worth correcting.

**10. Unused field `FLastMouseOverIndex`.**
Declared (`82`) and initialized to -1 (`472`), never read. Looks like a leftover, distinct from the pre-existing `FLastMouseMoveIndex`.

### Style and docs

**11. Convention deviations in the new code** (`.claude/rules/pascal.md`): pre-declared `var` blocks instead of inline variables; `begin` on its own line after `if`/`else`; a `//` comment in `MeasureItem` rather than `{ }`; and an over-indented `if` at `InvalidateCheck` (`1713`-`1714`).

martijnlaan:
I have cherry picked 0da22222 into main, and credited you in whatsnew. Let me know if you don't want to be credited by name there.

leserg73:
(_Translated with DeepL.com_)
Thank you for your feedback.
But you wrote so much that I need some time to review it.
The current implementation of the list view as a tree is the simplest approach and is limited by TCustomListBox, which is based on the standard Windows ListBox. For hidden items, the height is set to 1 pixel, and when there are a large number of child items, there will be a large gap between adjacent parents.

For a small number of child elements (5–10), this drawback can be ignored. But with a sufficiently large number, a different implementation is needed. Although in practice, I haven’t encountered huge lists of components.

> I have cherry picked 0da2222 into main, and credited you in whatsnew. Let me know if you don't want to be credited by name there.

Thanks.
I don't mind.

martijnlaan:
Before you put more time into it: I'm sorry to say, but I don't think the 1-height approach is sound, and I don't think you should continue with it. The added complexity is only worth it if it can handle large component lists, and the 1-height approach never will. To do it properly the items would have to be really removed from the list, while still updating item state in memory and keeping it available to callers by index. For that reason I'll close this pull request, sorry. Thanks for the effort you put into it though.

About the color properties I have similar thoughts: color properties are ignored by VCL styling, so it's odd to honor them in just this one place. But a separate, clean PR for those might be mergeable. Same for the MouseOver support. Please ensure all commits are signed, like your initial ones.

PR #600, opened 2026-06-22
ChineseTraditional.isl: Address remaining #599 review notes and further consistency fixes

GoneTone:
## Summary

Follow-up to #599. This addresses the remaining review notes from that PR, and also includes additional consistency fixes found while reviewing the whole file against Default.isl.

## Issues from the #599 review

- LangOptions: use Microsoft JhengHei for DialogFontName and WelcomeFontName instead of Arial and Segoe UI, which lack Traditional Chinese glyphs.
- ExtractingLabel: use 解壓縮檔案 to match StatusExtractFiles for the identical English "Extracting files...".
- FinishedHeadingLabel: include [name] to reflect "Completing the [name] Setup Wizard".
- SetupAborted: render "run Setup again" as 再次執行安裝程式, matching PreviousInstallNotCompleted and PrepareToInstallNeedsRestart.
- ErrorReadingExistingDest: use 現有檔案 for "the existing file", matching the surrounding messages.
- ErrorRegSvr32Failed: use 結束代碼 for "exit code".
- UninstallUnsupportedVer: remove the spaces around 「%1」 to match the other filename references.
- ConfirmUninstall: render "all of its components" as 及其所有元件.
- UninstalledMost: render "Some elements could not be removed. These can be removed manually." as 某些元件無法移除，您可以自行移除這些元件.

## Additional consistency fixes

- StatusRegisterFiles: use 註冊 for "register" (matching ErrorRegisterServer and ErrorRegisterTypeLib), instead of 登錄 which is used for "registry".
- StatusSavingUninstall: add the 正在 prefix used by the other status messages.
- ApplicationsFound2: render "Setup" as 安裝程式 (not 本安裝程式) and "the applications" as 這些應用程式 (plural).
- InstallingLabel: add the trailing period present in the English and in FinishedLabelNoIcons.
- InformationTitle / WizardInfoBefore / WizardInfoAfter: use 資訊 for "Information".
- ErrorProgress: render "%1 of %2" as %1 / %2 instead of the ambiguous %1 之 %2.
- Section comments: revert two comments that had been translated ("wizard page and", "Archive extraction failure details") back to English to match the rest, restore "stuff" in the "Setup Needs the Next Disk" comment, and remove two trailing spaces not present in the English.

PR #601, opened 2026-06-22
ChineseTraditional.isl: Address remaining #600 review notes and further consistency fixes

GoneTone:
## Summary

Follow-up to #600. This addresses the remaining review notes from that PR, and also includes a few more consistency and faithfulness fixes found while reviewing the whole file against Default.isl.

## Issues from the #600 review

- LangOptions: set DialogFontSize back to the Default.isl default of 9; only the font face is changed (to Microsoft JhengHei), which Traditional Chinese needs.
- PowerUserPrivilegesRequired: render "as a member of the Power Users group" with the group concept (Power Users 群組的成員) instead of dropping it and using the singular "Power User".
- ExitSetupMessage: render "to complete the installation" as 以完成安裝 rather than 以完成安裝程式 (the Setup program).
- StatusCreateIniEntries: use 建立 for "Creating" and 建立 INI 項目, matching ErrorIniEntry and the other status messages (previously 寫入).
- StatusCreateRegistryEntries: use 建立 for "Creating" and 登錄項目 for "registry entries" (previously 更新系統登錄).
- UninstallStatusLabel: end with a period to match the English, instead of "...".

## Additional consistency fixes

- PrivilegesRequiredOverrideText1 / PrivilegesRequiredOverrideText2: render "administrative privileges" as 系統管理員權限, matching OnlyAdminCanUninstall and the 系統管理員 wording used elsewhere.
- DiskSpaceWarning: render "continue anyway" as 仍要繼續安裝 to keep the "anyway" sense, matching NoUninstallWarning and AddonHostProgramNotFound.
- ConfirmDeleteSharedFile2: use the singular for "this shared file" / "it" / "the file", matching the English, which refers to a single file.


martijnlaan:
@GoneTone Refined my review skill, could you look into these new findings? Thanks!

1. Line 58 (SetupFileCorruptOrWrongVer): the value 請重新取得此軟體 renders only
   the second of the two remedies in the English. The English is "The setup files
   are corrupted, or are incompatible with this version of Setup. Please correct
   the problem or obtain a new copy of the program." The sibling message on line 56
   (SetupFileMissing) renders the same instruction in full as
   請修正此問題或重新取得此軟體。
2. Lines 70 and 71 (SetupAppRunningError, UninstallAppRunningError): the value
   請關閉該程式 renders "all instances of it" as "the program". The English is
   "Please close all instances of it now, then click OK to continue, or Cancel to
   exit."
3. Line 171 (BadDirName32): the value 資料夾名稱不得包含以下特殊字元 adds 特殊
   ("special"), which is not in the English "Folder names cannot include any of the
   following characters:". The sibling message on line 205 (BadGroupName) renders
   the same phrase as 資料夾名稱不得包含下列字元 without 特殊.
4. Lines 346 and 352 (ErrorChangingAttr, ErrorReplacingExistingFile): the qualifier
   "existing" is dropped. The English is "An error occurred while trying to change
   the attributes of the existing file:" and "An error occurred while trying to
   replace the existing file:". The related message on line 335
   (ErrorReadingExistingDest) renders "the existing file" as 現有檔案.
5. Line 383 (UninstalledMost): the English "elements" is rendered 元件, the term
   used elsewhere for "components" (the Select Components page). The English is
   "%1 uninstall complete.%n%nSome elements could not be removed. These can be
   removed manually."

GoneTone:
@martijnlaan PR: https://github.com/jrsoftware/issrc/pull/604

PR #606, opened 2026-06-27
Set dark scrollbar

leserg73:
Set a dark color for the scrollbar in dark mode for the CHM Help.

In a CHM help file with a dark theme, the scrollbars use standard coloring:
CHM-Dark_before

After applying the proposed patch, the scrollbar colors will match the dark theme:
CHM-Dark_after


martijnlaan:
Thanks! See 6e242640814a2826f5584920df10171d702e46e9 for the merged version: I tested it on Windows 7 and Windows 8 as well and for Windows 7 it needs `arrow-color` to be different from `face-color`. Also tweaked the other colors a bit.

PR #609, opened 2026-07-01
Update Italian.isl

bovirus:
@martijnlaan 

Update Italian language.

Are two fixes (missing"." at the end of second period).

Note: for this long string multi period

PrepareToInstallNeedsRestart=Il programma di installazione deve riavviare il computer. Per completare l'installazione di [name] dopo aver riavviato il sistema esegui di nuovo il programma di installazione.%n%nVuoi riavviare il sistema?

could be useful separate first and second period

PrepareToInstallNeedsRestart=Il programma di installazione deve riavviare il computer.**%n**Per completare l'installazione di [name] dopo aver riavviato il sistema esegui di nuovo il programma di installazione.%n%nVuoi riavviare il sistema?

Thanks.

### DIFF
--- a/Files/Languages/Italian.isl
+++ b/Files/Languages/Italian.isl
@@ -9,7 +9,7 @@
 ; Setup adds the periods automatically (appending a period would result in
 ; two periods being displayed).
 ;
-; Italian.isl - Last Update: 01.07.2026  by bovirus (bovirus@gmail.com)
+; Italian.isl - Last Update: 01.07.2026 by bovirus (bovirus@gmail.com)
 ;
 ; Translator name:   bovirus
 ; Translator e-mail: bovirus@gmail.com
--- a/Files/Languages/Italian.isl
+++ b/Files/Languages/Italian.isl
@@ -9,7 +9,7 @@
 ; Setup adds the periods automatically (appending a period would result in
 ; two periods being displayed).
 ;
-; Italian.isl - Last Update: 02.05.2026  by bovirus (bovirus@gmail.com)
+; Italian.isl - Last Update: 01.07.2026  by bovirus (bovirus@gmail.com)
 ;
 ; Translator name:   bovirus
 ; Translator e-mail: bovirus@gmail.com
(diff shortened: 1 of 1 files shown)


bovirus:
@martijnlaan 

Thanks.
Are you agree if I change

PrepareToInstallNeedsRestart=Il programma di installazione deve riavviare il computer.%nPer completare l'installazione di [name] dopo aver riavviato il sistema esegui di nuovo il programma di installazione.%n%nVuoi riavviare il sistema?

adding the newline?

martijnlaan:
Thanks. I merged it, but see now this only introduces more issues and should be reverted. These sentences should not have periods, see `Default.isl`. Also please don't add line breaks, especially not unpaired ones. All line breaks you added in earlier versions should be removed. Translations should always be as faithful as possible.

Please take a look at these update proposals:
Update 1: https://github.com/jrsoftware/issrc/commit/5445c8094d0097046a9ef0293dab4be39e7a6c44
Update 2: https://github.com/jrsoftware/issrc/commit/9b18bd9e92bc27640eae2ab45d71f1a8eb9643cc
(Mini) update 3: https://github.com/jrsoftware/issrc/commit/a4a3344da01ee398ff00546d177ec80f7acfbe86

First it lists the issues it is trying to fix, and then the proposed changes are shown.

bovirus:
@martijnlaan

Thanks.

I see that in default.isl was missing point at the end. I considered that was an error.
Because these are two separate periods and second period has missing "." at the end,

About long string please consider that you have  double %n just after first part of long string (with two period) without newline.
My propsoal is to add a %n to separate the first two period and remove one of double %n after.

My idea is to have when you can each period separate by newline for beter readibility.

I will wait for the fixes.

martijnlaan:
These periods are added automatically at runtime, and shouldn't be in the `.isl`.

Suggestions to improve `Default.isl` are welcome, but please do not apply them already to your own translation. Will you look into the three update proposals and tell me if I can apply them to the real `Italian.isl`?

bovirus:
@martijnlaan

Coudl be usefull for the strings without period at the end to add a note to avoid that someone make the same changes that I did.

martijnlaan:
Thanks for the suggestion, but such a note is already there at the top of `Default.isl`, and always has been there.

PR #621, opened 2026-07-22
Update Italian.isl

bovirus:
@martijnlaan 

Update Italian language following some feedbacks received about 

- removing "," not necessary for the Italian flow in the specific context
- change Italian string flow regarding using the Italian "per xxx" 

Thanks.

### DIFF
--- a/Files/Languages/Italian.isl
+++ b/Files/Languages/Italian.isl
@@ -67,8 +67,8 @@ WinVersionTooLowError=Questo programma richiede %1 versione %2 o successiva.
 WinVersionTooHighError=Questo programma non può essere installato su %1 versione %2 o successiva.
 AdminPrivilegesRequired=Per installare questo programma sono richiesti privilegi di amministratore.
 PowerUserPrivilegesRequired=Per poter installare questo programma sono richiesti i privilegi di amministratore o di Power Users.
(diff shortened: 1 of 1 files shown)


martijnlaan:
Thank you

martijnlaan:
FYI, fixed two stray spaces introduced by this commit, see e01abf47

martijnlaan:
Did another fix: In SetupAppRunningError and UninstallAppRunningError the "per continuare" got lost in this commit, so "per uscire" now covered both buttons and the message told the user OK exits too. Restored as seleziona "OK" per continuare o "Annulla" per uscire. ClickNext and ReadyLabel2a are fine, there the fronted "Per continuare" kept both purposes.

Also happened to notice both messages were missing the "Setup/Uninstall has detected that" text, which is also why the two translations were identical. Added it back. NoUninstallWarning was missing the same opening, fixed too.

See 21035bb7

bovirus:
@martijnlaan 

Very strange so many errors. Expeacially misisng starting "Setup/Uninstall has detected that" because is not touched by my changes that are regarding the end of the string. 
I got original and changed it directly.
So sorry for the issue.
I beeiev that i need holiday (I'm too tired)... ;-)

martijnlaan:
No problem and thanks for checking. I beleive the "Setup/Uninstall has detected that" texts have been missing since forever.

bovirus:
@martijnlaan 

Thanks again.

PR #622, opened 2026-07-23
Add Zstandard compression support

amyspark:
hi! 

Following up on earlier inquiries on the mailing thread, I'm filing this proposal to add Zstd support to Inno's installer payload. 

My use case is to improve the performance of solid compressed installers; in my local testing (~600MB payload, comprising some 5GB worth of headers/DLLs/PDBs) I got a 2x speedup when skipping payloads, from 56s (with `lzma2/max`) to 24s (with `zstd/18`). With non-solid payloads the improvements are much more nuanced, from 22s down to 19s. 

The other benefit that libzstd brings is to use automatically managed multithreaded compression, without the need for glue code; this is especially useful for my use case since I rely on high compression levels. In this case, each compression call will cause data to be ingested automatically, and require the IS compiler to wait and write the payload on `TZstdCompressor.DoFinish` as it becomes available.

The new `Zstd` unit and associated glue code is based on the `Compiler.Zlib` implementation, with levels ranging from 1 to 22 (default = 0, which is managed by libzstd). Since I could not pass a number of worker threads (more on that below),  I hodge podged autodetected thread count detection, which I set to half the number of processors.

TODO / Questions / Need for improvement
--

- Would it be possible to subclass `TCompressorProps` so as to allow Zstd-specific parameters to be passed? There does is support, see https://github.com/jrsoftware/issrc/blob/67ccc36c179d4be50e6775c1d757d0705d5b18a3/Projects/Src/Compression.LZMACompressor.pas#L914-L916 but https://github.com/jrsoftware/issrc/blob/67ccc36c179d4be50e6775c1d757d0705d5b18a3/Projects/Src/Compiler.SetupCompiler.pas#L7800-L7801 there is no provision for using a different Props class.

- Related to the earlier, would it be possible to pin the number of threads when `UseSolidCompression=yes` to 2? I think this is also done for LZMA(2) but can't say for sure. When not using solid compression, there is no point in using multiple threads as compression is done asynchronously; and in my use case, with very small (usually less than 5MB) files.

Thanks for your feedback!

### DIFF
--- a/Projects/Bin/synch-isfiles.bat
+++ b/Projects/Bin/synch-isfiles.bat
@@ -14,6 +14,10 @@ copy ..\..\Files\isscint*.dll
 copy ..\..\Files\isscint*.dll.issig
 copy ..\..\Files\islzma*.dll
 copy ..\..\Files\islzma*.dll.issig
+copy ..\..\Files\isunzstd*.dll
+copy ..\..\Files\isunzstd*.dll.issig
+copy ..\..\Files\iszstd*.dll
+copy ..\..\Files\iszstd*.dll.issig
 copy ..\..\Files\islzma*.exe
 copy ..\..\Files\islzma*.exe.issig
 copy ..\..\Files\WizClassicSmallImage.bmp
--- a/Projects/ISCmplr.dproj
+++ b/Projects/ISCmplr.dproj
@@ -134,6 +134,7 @@
         <DCCReference Include="Src\Compression.Zlib.pas"/>
         <DCCReference Include="Src\Compression.bzlib.pas"/>
         <DCCReference Include="Src\Compression.LZMACompressor.pas"/>
+        <DCCReference Include="Src\Compression.Zstd.pas"/>
         <DCCReference Include="Src\Shared.FileClass.pas"/>
         <DCCReference Include="..\Components\ChaCha20.pas"/>
         <DCCReference Include="Src\Shared.VerInfoFunc.pas"/>
--- a/Projects/Src/Compiler.Messages.pas
+++ b/Projects/Src/Compiler.Messages.pas
@@ -114,6 +114,7 @@ const
   SCompilerArchitectureIdentifierDeprecatedWarning = 'Architecture identifier "%s" is deprecated. ' +
     'Substituting "%s", but note that "%s" is preferred in most cases. See the "Architecture Identifiers" topic in help file for more information.';
   SCompilerRegTypeLibArchitectureMismatch = '%d-bit Setup cannot register type libraries in the %d-bit registry view. If you intend for the type library to be registered in the %0:d-bit registry view, add the "%0:dbit" flag';
+  SCompilerZstdTooManyThreads = 'CompressionThreads should be set to a fixed number when using Zstd compression with SolidCompression=no. More than one thread may lead to poor compression speeds with small files.';
 
   { Signing }
   SCompilerSignatureNeeded = 'Signed uninstaller mode is enabled. Using ' +
--- a/Projects/Src/Compiler.SetupCompiler.pas
+++ b/Projects/Src/Compiler.SetupCompiler.pas
@@ -33,8 +33,8 @@ uses
   SimpleExpression, SHA256, ChaCha20, Shared.SetupTypes, Shared.CommonFunc,
   Shared.Struct, Shared.CompilerInt.Struct, Shared.PreprocInt, Shared.SetupMessageIDs,
   Shared.SetupSectionDirectives, Shared.VerInfoFunc, Shared.DebugStruct,
-  Compiler.ScriptCompiler, Compiler.StringLists, Compression.LZMACompressor,
-  Compiler.ExeUpdateFunc;
+  Compiler.ScriptCompiler, Compiler.StringLists, Compiler.ExeUpdateFunc,
+  Compression.LZMACompressor, Compression.Zstd;
 
 type
   EISCompileError = class(Exception);
@@ -74,7 +74,7 @@ type
 
   TCodeParameterKind = (cpkCheck, cpkDirectiveCheck, cpkInstall, cpkOnLog);
 
-  TPrecompiledFile = (pfSetup, pfSetupCustomStyle, pfSetupLdr, pfIs7z, pfIsbunzip, pfIsunzlib, pfIslzma);
+  TPrecompiledFile = (pfSetup, pfSetupCustomStyle, pfSetupLdr, pfIs7z, pfIsbunzip, pfIsunzlib, pfIslzma, pfIsunzstd);
   TPrecompiledFiles = set of TPrecompiledFile;
 
   TWizardImages = TObjectList<TCustomMemoryStream>;
@@ -246,6 +246,7 @@ type
     procedure InitLZMADLL;
     procedure InitPreprocessor;
     procedure InitZipDLL;
+    procedure InitZstdDLL;
     procedure PopulateLanguageEntryData;
     procedure ProcessMinVersionParameter(const ParamValue: TParamValue;
       var AMinVersion: TSetupVersionData);
@@ -374,7 +375,7 @@ type
   end;
 
 var
-  ZipInitialized, BzipInitialized, LZMAInitialized: Boolean;
+  ZipInitialized, BzipInitialized, LZMAInitialized, ZstdInitialized: Boolean;
   PreprocessorInitialized: Boolean;
   PreprocessScriptProc: TPreprocessScriptProc;
 
@@ -617,6 +618,34 @@ begin
   PreprocessorInitialized := True;
 end;
 
+procedure TSetupCompiler.InitZstdDLL;
+begin
+  if ZstdInitialized then
+    Exit;
+  var Filename: String;
+  {$IFDEF WIN64}
+  var DllName: String;
+  if IsArm64 then begin
+    { We can use an Arm64EC DLL from our x64 EXE, for better performance }
+    DllName := 'iszstd-Arm64EC.dll';
+    const Arm64Filename = CompilerDir + DllName;
+    if NewFileExists(Arm64Filename) then { Allow it to be deleted, for easy performance comparison }
+      Filename := Arm64Filename;
+  end;
+  if Filename = '' then begin
+    DllName := 'iszstd-x64.dll';
+    Filename := CompilerDir + DllName;
+  end;
+  {$ELSE}
+  const DllName = 'iszstd.dll';
+  Filename := CompilerDir + DllName;
+  {$ENDIF};
+  const M = LoadCompilerDLL(Filename, [ltloTrustAllOnDebug]);
+  if not ZstdInitCompressFunctions(M) then
+    AbortCompile('Failed to get address of functions in ' + DllName);
+  ZstdInitialized := True;
+end;
+
 procedure TSetupCompiler.InitZipDLL;
 begin
   if ZipInitialized then
@@ -642,40 +671,6 @@ begin
 end;
 
 procedure TSetupCompiler.InitLZMADLL;
-
-  {$IFDEF WIN64}
-  
-  function IsArm64: Boolean;
-  const
-    IMAGE_FILE_MACHINE_ARM64 = $AA64;
-    {$IFNDEF CPUX64}
-    PROCESSOR_ARCHITECTURE_ARM64 = 12;
-    {$ENDIF}
-  var
-    IsWow64Process2Func: function(hProcess: THandle; var pProcessMachine, pNativeMachine: USHORT): BOOL; stdcall;
-  begin
-    const KernelModule = GetModuleHandle(kernel32);
-
-    IsWow64Process2Func := GetProcAddress(KernelModule, 'IsWow64Process2');
-    var ProcessMachine, NativeMachine: USHORT;
-    if Assigned(IsWow64Process2Func) and
-       IsWow64Process2Func(GetCurrentProcess, ProcessMachine, NativeMachine) then
-      Exit(NativeMachine = IMAGE_FILE_MACHINE_ARM64);
-
-    { When running with x64 emulation on ARM64, GetNativeSystemInfo will just lie to us, so only
-      call if not x64 (which currently is impossible) }
-    {$IFNDEF CPUX64}
-    var SysInfo: TSystemInfo;
-    GetNativeSystemInfo(SysInfo);
-    if SysInfo.wProcessorArchitecture = PROCESSOR_ARCHITECTURE_ARM64 then
-      Exit(True);
-    {$ENDIF}
-
-    Result := False;
-  end;
-  
-  {$ENDIF}
-
 begin
   if LZMAInitialized then
     Exit;
@@ -2679,7 +2674,7 @@ var
   function StrToPrecompiledFiles(S: String): TPrecompiledFiles;
   const
     PrecompiledFiles: array of PChar = ['setup', 'setupcustomstyle', 'setupldr',
-      'is7z', 'isbunzip', 'isunzlib', 'islzma'];
+      'is7z', 'isbunzip', 'isunzlib', 'islzma', 'isunzstd'];
   begin
     Result := [];
     while True do
@@ -2693,6 +2688,7 @@ var
         4: Include(Result, pfIsbunzip);
         5: Include(Result, pfIsunzlib);
         6: Include(Result, pfIslzma);
+        7: Include(Result, pfIsunzstd);
       end;
   end;
 
@@ -2941,6 +2937,10 @@ begin
           CompressMethod := cmLZMA2;
           CompressLevel := clLZMAMax;
         end
+        else if Value = 'zstd' then begin
+          CompressMethod := cmZstd;
+          CompressLevel := 18;
+        end
         else if Copy(Value, 1, 4) = 'zip/' then begin
           I := StrToIntDef(Copy(Value, 5, Maxint), -1);
           if (I < 1) or (I > 9) then
@@ -2965,6 +2965,13 @@ begin
           if not LZMAGetLevel(Copy(Value, 7, Maxint), I) then
             Invalid;
           CompressMethod := cmLZMA2;
+          CompressLevel := I;
+        end
+        else if Copy(Value, 1, 5) = 'zstd/' then begin
+          I := StrToIntDef(Copy(Value, 6, Maxint), -1);
+          if (I < 1) or (I > 22) then
+            Invalid;
+          CompressMethod := cmZstd;
           CompressLevel := I;
         end
         else
@@ -7431,7 +7438,7 @@ var
       WriteWizardImages(WizardImagesDynamicDark, W, WizardImages);
       WriteWizardImages(WizardSmallImagesDynamicDark, W, WizardSmallImages);
       WriteWizardImages(WizardBackImagesDynamicDark, W, WizardBackImages);
-      if SetupHeader.CompressMethod in [cmZip, cmBzip] then
+      if SetupHeader.CompressMethod in [cmZip, cmBzip, cmZstd] then
         WriteStream(DecompressorDLL, W);
       if SevenZipDLL <> nil then
         WriteStream(SevenZipDLL, W);
@@ -7613,6 +7620,10 @@ var
             end;
           cmLZMA2: begin
               Result := TLZMA2Compressor;
+            end;
+          cmZstd: begin
+              InitZstdDLL;
+              Result := TZSTDCompressor;
             end;
         else
           AbortCompile('GetCompressorClass: Unknown CompressMethod');
@@ -8372,6 +8383,10 @@ begin
     CallIdleProc;
 
     { Verify settings set in [Setup] section }
+    if CompressMethod = cmZstd then begin
+      if (CompressProps.NumBlockThreads = 0) and (not UseSolidCompression) then
+        WarningsList.Add(SCompilerZstdTooManyThreads);
+    end;
     if SetupDirectiveLines[ssUseSetupLdr] = 0 then begin
       if SetupArchitecture = sa32bit then
         SetupLdrArchitecture := sla32bit
@@ -8982,6 +8997,12 @@ begin
           AddStatus(Format(SCompilerStatusReadingFile, [DllName]));
           DecompressorDLL := CreateMemoryStreamFromFile(CompilerDir + DllName,
             not(pfIsbunzip in DisablePrecompiledFileVerifications), OnCheckedTrust);
+        end;
+      cmZstd: begin
+          const DllName = Format('isunzstd%s.dll', [DllNameExtension]);
+          AddStatus(Format(SCompilerStatusReadingFile, [DllName]));
+          DecompressorDLL := CreateMemoryStreamFromFile(CompilerDir + DllName,
+            not(pfIsunzstd in DisablePrecompiledFileVerifications), OnCheckedTrust);
         end;
     end;
 
--- a/Projects/Src/Compression.Zstd.pas
+++ b/Projects/Src/Compression.Zstd.pas
@@ -1,0 +1,373 @@
+unit Compression.Zstd;
+
+{
+  Inno Setup
+  Copyright (C) 1997-2026 Jordan Russell
+  Portions by Martijn Laan
+  For conditions of distribution and use, see LICENSE.TXT.
+
+  Originally contributed by Amyspark <amy@amyspark.me>
+
+  Declarations for Zstandard functions & structures
+}
+
+interface
+
+uses
+  Windows, SysUtils, Compression.Base;
+
+function ZstdInitCompressFunctions(Module: HMODULE): Boolean;
+function ZstdInitDecompressFunctions(Module: HMODULE): Boolean;
+
+type
+  TZSTD_inBuffer = record
+    src: Pointer;        { start of input buffer }
+    size: NativeUInt;        { size of input buffer }
+    pos: NativeUInt;         { position where reading stopped. Will be updated. Necessarily 0 <= pos <= size }
+  end;
+
+  TZSTD_outBuffer = record
+    dst: Pointer;        { start of output buffer }
+    size: NativeUInt;        { size of output buffer }
+    pos: NativeUInt;         { position where writing stopped. Will be updated. Necessarily 0 <= pos <= size }
+  end;
+
+  TZSTD_frameProgression = record
+    ingested: UInt64;   { nb input bytes read and buffered }
+    consumed: UInt64;   { nb input bytes actually compressed }
+    produced: UInt64;   { nb of compressed bytes generated and buffered }
+    flushed: UInt64;    { nb of compressed bytes flushed : not provided; can be tracked from caller side }
+    currentJobID: Cardinal;         { MT only : latest started job nb }
+    nbActiveWorkers: Cardinal;      { MT only : nb of workers actively compressing at probe time }
+  end;
+
+  TZstdCompressor = class(TCustomCompressor)
+  private
+    FCompressionLevel: Integer;
+    FNumThreads: Integer;
+    FInitialized: Boolean;
+    FStrm: Pointer;
+    FOut: TZSTD_outBuffer;
+    FBuffer: array[0..$FFFFF] of Byte;
+    { Workaround for Zstd not resetting the frame progression until compress2
+      is called. Let's keep a good local copy }
+    FProgress: TZSTD_FrameProgression;
+    procedure EndCompress;
+    procedure FlushBuffer;
+    procedure InitCompress;
+  protected
+    procedure DoCompress(const Buffer; Count: Cardinal); override;
+    procedure DoFinish; override;
+  public
+    constructor Create(AWriteProc: TCompressorWriteProc;
+      AProgressProc: TCompressorProgressProc; CompressionLevel: Integer;
+      ACompressorProps: TCompressorProps); override;
+    destructor Destroy; override;
+  end;
+
+  TZstdDecompressor = class(TCustomDecompressor)
+  private
+    FInitialized: Boolean;
+    FStrm: Pointer;
+    FIn: TZSTD_inBuffer;
+    FReachedEnd: Boolean;
+    FBuffer: array[0..$FFFFF] of Byte;
+  public
+    constructor Create(AReadProc: TDecompressorReadProc); override;
+    destructor Destroy; override;
+    procedure DecompressInto(var Buffer; Count: Cardinal); override;
+    procedure Reset; override;
+  end;
+
+implementation
+
+const
+  SZstdDataError = 'zstd: Compressed data is corrupted';
+  SZstdInternalError = 'zstd: Internal error. Code %d';
+
+  ZSTD_e_continue = 0;
+  ZSTD_e_flush    = 1;
+  ZSTD_e_end      = 2;
+
+  ZSTD_c_nbWorkers        = 400;
+
+  ZSTD_reset_session_only = 1;
+
+var
+  ZSTD_createCStream: function: Pointer; stdcall;
+  ZSTD_initCStream: function (zcs: Pointer; compressionLevel: Integer): NativeUInt; stdcall;
+  ZSTD_CCtx_setParameter: function(cctx: Pointer; param: Cardinal; value: Integer): NativeUInt; stdcall;
+  ZSTD_compressStream2: function(cctx: Pointer; var output: TZSTD_outBuffer; var input: TZSTD_inBuffer; endOp: Cardinal): NativeUInt; stdcall;
+  ZSTD_freeCStream: function(zcs: Pointer): NativeUInt; stdcall;
+
+  ZSTD_createDStream: function(): Pointer; stdcall;
+  ZSTD_initDStream: function(zds: Pointer): NativeUInt; stdcall;
+  ZSTD_decompressStream: function(zds: Pointer; var output: TZSTD_outBuffer; var input: TZSTD_inBuffer): NativeUInt; stdcall;
+  ZSTD_freeDStream: function(zds: Pointer): NativeUInt; stdcall;
+
+  ZSTD_isError: function(res: NativeUInt): Cardinal; stdcall;
+  ZSTD_CCtx_reset: function(cctx: Pointer; reset: Cardinal): NativeUInt; stdcall;
+  ZSTD_getFrameProgression: function(cctx: Pointer): TZSTD_frameProgression; stdcall;
+
+function ZstdInitCompressFunctions(Module: HMODULE): Boolean;
+begin
+  ZSTD_createCStream := GetProcAddress(Module, 'ZSTD_createCStream');
+  ZSTD_CCtx_setParameter := GetProcAddress(Module, 'ZSTD_CCtx_setParameter');
+  ZSTD_initCStream := GetProcAddress(Module, 'ZSTD_initCStream');
+  ZSTD_compressStream2 := GetProcAddress(Module, 'ZSTD_compressStream2');
+  ZSTD_freeCStream := GetProcAddress(Module, 'ZSTD_freeCStream');
+  ZSTD_isError := GetProcAddress(Module, 'ZSTD_isError');
+  ZSTD_CCtx_reset := GetProcAddress(Module, 'ZSTD_CCtx_reset');
+  ZSTD_getFrameProgression := GetProcAddress(Module, 'ZSTD_getFrameProgression');
+  Result := Assigned(ZSTD_createCStream) and Assigned(ZSTD_CCtx_setParameter)
+    and Assigned(ZSTD_initCStream) and Assigned(ZSTD_compressStream2)
+    and Assigned(ZSTD_freeCStream) and Assigned(ZSTD_isError)
+    and Assigned(ZSTD_CCtx_reset) and Assigned(ZSTD_getFrameProgression);
+  if not Result then begin
+    ZSTD_createCStream := nil;
+    ZSTD_initCStream := nil;
+    ZSTD_CCtx_setParameter := nil;
+    ZSTD_compressStream2 := nil;
+    ZSTD_freeCStream := nil;
+    ZSTD_isError := nil;
+    ZSTD_CCtx_reset := nil;
+    ZSTD_getFrameProgression := nil;
+  end;
+end;
+
+function ZstdInitDecompressFunctions(Module: HMODULE): Boolean;
+begin
+  ZSTD_createDStream := GetProcAddress(Module, 'ZSTD_createDStream');
+  ZSTD_initDStream := GetProcAddress(Module, 'ZSTD_initDStream');
+  ZSTD_decompressStream := GetProcAddress(Module, 'ZSTD_decompressStream');
+  ZSTD_freeDStream := GetProcAddress(Module, 'ZSTD_freeDStream');
+  ZSTD_isError := GetProcAddress(Module, 'ZSTD_isError');
+  Result := Assigned(ZSTD_createDStream) and Assigned(ZSTD_initDStream) and
+    Assigned(ZSTD_decompressStream) and Assigned(ZSTD_freeDStream) and Assigned(ZSTD_isError);
+  if not Result then begin
+    ZSTD_createDStream := nil;
+    ZSTD_initDStream := nil;
+    ZSTD_decompressStream := nil;
+    ZSTD_freeDStream := nil;
+    ZSTD_isError := nil;
+  end;
+end;
+
+function Check(const Code: NativeUInt): Boolean;
+begin
+  if ZSTD_isError(Code) <> 0 then
+    raise ECompressInternalError.CreateFmt(SZstdInternalError, [Code]);
+  Result := True;
+end;
+
+{ TZstdCompressor }
+
+constructor TZstdCompressor.Create(AWriteProc: TCompressorWriteProc;
+  AProgressProc: TCompressorProgressProc; CompressionLevel: Integer;
+  ACompressorProps: TCompressorProps);
+var
+  GetActiveProcessorCountFunc: function(GroupNumber: WORD): WORD; stdcall;
+begin
+  inherited;
+  FCompressionLevel := CompressionLevel;
+  FNumThreads := 0;
+  if ACompressorProps is TThreadedCompressorProps then begin
+    const Props = (ACompressorProps as TThreadedCompressorProps);
+    if Props.NumBlockThreads <> 0 then
+      FNumThreads := Props.NumBlockThreads;
+  end;
+  if FNumThreads = 0 then begin
+    { Let's make Zstd use automatically all physical processors }
+    GetActiveProcessorCountFunc := GetProcAddress(GetModuleHandle(kernel32),
+      'GetActiveProcessorCount');
+    if Assigned(GetActiveProcessorCountFunc) then begin
+      const ActiveProcessorCount = GetActiveProcessorCountFunc(65535);
+      if ActiveProcessorCount > 1 then
+        FNumThreads := ActiveProcessorCount div 2;
+    end;
+  end;
+  InitCompress;
+end;
+
+destructor TZstdCompressor.Destroy;
+begin
+  { Unlike other destructors e.g. Zlib, this library is backed by a runtime
+    thread pool. For this reason the buffer needs to be as small as
+    possible (1 block == 128KB + overhead), so as to trigger the flush
+    ASAP, then every write afterwards is promptly discarded from memory.
+    This helps a lot with returning the control to the event loop, minimizing
+    the time the UI spends unresponsive. }
+  var FUnusedIn: TZSTD_inBuffer;
+  FillChar(FUnusedIn, SizeOf(FUnusedIn), 0);
+  FOut.pos := 0;
+  var OldCount := FProgress.consumed;
+  var Code: NativeUInt := 1;
+  if Assigned(ProgressProc) and (FProgress.consumed <> FProgress.ingested) then begin
+    While Code <> 0 do begin
+      Code := ZSTD_compressStream2(FStrm, FOut, FUnusedIn, ZSTD_e_flush);
+      Check(Code);
+      FProgress := ZSTD_getFrameProgression(FStrm);
+      const Actual = FProgress.consumed - OldCount;
+      ProgressProc(Actual);
+      OldCount := FProgress.consumed;
+      FOut.pos := 0;
+    end;
+  end;
+  EndCompress;
+  Check(ZSTD_freeCStream(FStrm));
+  inherited;
+end;
+
+procedure TZstdCompressor.InitCompress;
+begin
+  { Decoupling initialization from compression context creation allows
+    reusing the context for further compression operations. Also, in
+    multithreaded mode, it's pretty easy to OOM Delphi by using Zstd together
+    with SolidCompression=no, as Zstd allocates a thread pool per context }
+  if FStrm = nil then begin
+    FStrm := ZSTD_createCStream();
+    if FStrm = nil then
+      OutOfMemoryError;
+    Check(ZSTD_initCStream(FStrm, FCompressionLevel));
+    if FNumThreads > 1 then begin
+      const Code = ZSTD_CCtx_setParameter(FStrm, ZSTD_c_nbWorkers, FNumThreads);
+      if ZSTD_isError(Code) <> 0 then
+        Check(ZSTD_CCtx_setParameter(FStrm, ZSTD_c_nbWorkers, 1));
+    end;
+  end;
+  if not FInitialized then begin
+    FillChar(FProgress, SizeOf(FProgress), 0);
+    FillChar(FOut, SizeOf(FOut), 0);
+    FOut.dst := @FBuffer;
+    FOut.size := SizeOf(FBuffer);
+    FOut.pos := 0;
+    FInitialized := True;
+  end;
+end;
+
+procedure TZstdCompressor.EndCompress;
+begin
+  if FInitialized then begin
+    FInitialized := False;
+    { Only reset the compression state; the rest is reusable }
+    Check(ZSTD_CCtx_reset(FStrm, ZSTD_reset_session_only));
+  end;
+end;
+
+procedure TZstdCompressor.FlushBuffer;
+begin
+  if FOut.pos > 0 then begin
+    WriteProc(FBuffer, FOut.pos);
+    FOut.pos := 0;
+  end;
+end;
+
+procedure TZstdCompressor.DoCompress(const Buffer; Count: Cardinal);
+begin
+  InitCompress;
+  var LIn: TZSTD_inBuffer;
+  LIn.src := @Buffer;
+  LIn.size := Count;
+  LIn.pos := 0;
+  var OldCount := FProgress.consumed;
+  while LIn.pos < Count do begin
+    Check(ZSTD_compressStream2(FStrm, FOut, LIn, ZSTD_e_continue));
+    if FOut.pos = FOut.size then
+      FlushBuffer;
+    { Maximize responsiveness by tying ProgressProc to the actual data
+      ingested; especially helpful with compression levels >= 19 }
+    if Assigned(ProgressProc) then begin
+      FProgress := ZSTD_getFrameProgression(FStrm);
+      const Actual = FProgress.consumed - OldCount;
+      ProgressProc(Actual);
+      OldCount := FProgress.consumed;
+    end;
+  end;
+end;
+
+procedure TZstdCompressor.DoFinish;
+begin
+  var LIn: TZSTD_inBuffer;
+  InitCompress;
+  var LReachedEnd := False;
+  FillChar(LIn, SizeOf(LIn), 0);
+  var OldCount := FProgress.consumed;
+  while not LReachedEnd do begin
+    const Code = ZSTD_compressStream2(FStrm, FOut, LIn, ZSTD_e_end);
+    Check(Code);
+    FlushBuffer;
+    if Code = 0 then
+      LReachedEnd := True;
+    if Assigned(ProgressProc) then begin
+      FProgress := ZSTD_getFrameProgression(FStrm);
+      const Actual = FProgress.consumed - OldCount;
+      ProgressProc(Actual);
+      OldCount := FProgress.consumed;
+    end;
+  end;
+  EndCompress;
+end;
+
+{ TZstdDecompressor }
+
+constructor TZstdDecompressor.Create(AReadProc: TDecompressorReadProc);
+begin
+  inherited Create(AReadProc);
+  FStrm := ZSTD_createDStream();
+  if FStrm = nil then
+    OutOfMemoryError;
+  try
+    Reset;
+    FInitialized := True;
+  except
+    begin
+      Check(ZSTD_freeDStream(FStrm));
+      raise;
+    end;
+  end;
+end;
+
+destructor TZstdDecompressor.Destroy;
+begin
+  if FInitialized then begin
+    Check(ZSTD_freeDStream(FStrm));
+    FInitialized := False;
+  end;
+  inherited Destroy;
+end;
+
+procedure TZstdDecompressor.DecompressInto(var Buffer; Count: Cardinal);
+begin
+  var LOut: TZSTD_outBuffer;
+  LOut.dst := @Buffer;
+  LOut.size := Count;
+  LOut.pos := 0;
+  while LOut.pos < Count do begin
+    if FReachedEnd then  { unexpected EOF }
+      raise ECompressDataError.Create(SZstdDataError);
+    if FIn.pos = FIn.size then begin
+      FIn.src := @FBuffer;
+      FIn.size := ReadProc(FBuffer, SizeOf(FBuffer));
+      FIn.pos := 0;
+    end;
+    const OldInPos = FIn.pos;
+    const OldOutPos = LOut.pos;
+    const Code = ZSTD_decompressStream(FStrm, LOut, FIn);
+    if ZSTD_isError(Code) <> 0 then begin
+      raise ECompressDataError.Create(SZstdDataError);
+    end else if (FIn.pos = OldInPos) and (LOut.pos = OldOutPos) then begin
+      { Sanity check; no data consumed or compressed at all }
+      raise ECompressDataError.Create(SZstdDataError);
+    end else if Code = 0 then
+      FReachedEnd := True;
+  end;
+end;
+
+procedure TZstdDecompressor.Reset;
+begin
+  FillChar(FIn, SizeOf(FIn), 0);
+  Check(ZSTD_initDStream(FStrm));
+  FReachedEnd := False;
+end;
+
+end.
--- a/Projects/Src/IDE.ScintStylerInnoSetup.pas
+++ b/Projects/Src/IDE.ScintStylerInnoSetup.pas
@@ -2163,6 +2163,7 @@ end;
 
 type
   TZipLevel = 1..9;
+  TZstdLevel = 1..22;
 
 const
   LZMALevels: TArray<TScintRawString> = ['fast', 'normal', 'max', 'ultra', 'ultra64'];
@@ -2178,10 +2179,12 @@ function GetCompressionValues: TArray<TScintRawString>;
 const
   ZipAlgos: TArray<TScintRawString> = ['zip', 'bzip'];
   LZMAAlgos: TArray<TScintRawString> = ['lzma', 'lzma2'];
+  ZstdAlgo: String = 'zstd';
 begin
   SetLength(Result, 1 +
     Length(ZipAlgos) + Length(ZipAlgos) * (High(TZipLevel) - Low(TZipLevel) + 1) +
-    Length(LZMAAlgos) + Length(LZMAAlgos) * Length(LZMALevels));
+    Length(LZMAAlgos) + Length(LZMAAlgos) * Length(LZMALevels) +
+    1 + (High(TZstdLevel) - Low(TZstdLevel) + 1));
   var I := 0;
   SetResult(I, 'none');
   for var Algo in ZipAlgos do begin
@@ -2194,6 +2197,9 @@ begin
     for var Level in  LZMALevels do
       SetResult(I, TScintRawString(Algo + '/' + Level));
   end;
+  SetResult(I, TScintRawString(ZstdAlgo));
+  for var Level := Low(TZstdLevel) to High(TZstdLevel) do
+    SetResult(I, TScintRawString(ZstdAlgo + '/' + Level.ToString));
 end;
 
 initialization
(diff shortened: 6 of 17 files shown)


amyspark:
NB: I have set this PR to draft as I need to remove some components to make Inno buildable with Delphi 12.3 (see the first commit). Will remove that commit once we make progress!

martijnlaan:
Thank you. In general it looks to me like you did a good job with the integration, very nice 👍 I'm not familiar with zstd at all myself though. Also am not able to properly review the actual code 

Would like to know what @jordanrussell thinks about the principle of including it or not, and whether the actual class implementations in [Compression.Zstd.pas](https://github.com/jrsoftware/issrc/pull/622/changes/59c0a727ba5982b90fbb374c2ab301fd7f87401a#diff-4b320bc3fdc2c50c644f01cd179830e4a0d139becae4fad90a7ab2afbf724001) are like they are supposed to be.

What stands out most right now:
1. The zstd c code is in the wrong place, it should be in our [iscompress repo](https://github.com/jrsoftware/iscompress).
2. There should ideally be two DLLs and not one. `iszstd.dll` to compress, `isunzstd.dll` to decompress. Otherwise Setup will contain compression code.

Will you sign the CLA? See above. It would also be good if you could enable [verified commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification). Git comes with GPG now, so it's relatively easy to [set up](https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key).

amyspark:
> The zstd c code is in the wrong place, it should be in our [iscompress repo](https://github.com/jrsoftware/iscompress).

No problem, can I split the MR out once I've got this past review?

> There should ideally be two DLLs and not one. iszstd.dll to compress, isunzstd.dll to decompress. Otherwise Setup will contain compression code.

The push of a few moments ago fixes this (with a .def file + removing the automatic exports, for ease of review + parity with the upstream VS solution), plus adds the GPG signed commits as requested.

martijnlaan:
Thanks. I will look into downloading zstd sources into iscompress myself already, and also try to make the project files your created conform to the standard set by the zlib and bzlib project files. For example, the DLL's should be free of C runtime code. I also noticed both DLL's use all files still (as in: the decompression DLL references compression files). Just FYI, you shouldn't edit the project files yourself until further notice as it will just conflict with my edits.

martijnlaan:
Your PR includes dictBuilder files, but as I understand it, it's only needed to create a dictionary, and not the compress with it. Is that correct? Being able to remove dictBuilder would be good because cover.c is trying to display stuff to the console.

martijnlaan:
Spoke to Jordan and we would like to continue integrating this, so thank you for the proposal and the legwork. 

There's one issue: `Compression.Zstd.pas` currently has your copyright notice and is MPL licensed. For a file in `Projects\Src ` it must instead use our standard Inno Setup header, like the one in `Compression.Zlib.pas`. That way it also no longer contradicts the CLA. The header can and should still credit your work, for example with a line like 'Originally contributed by Amy Spark', or something similar. Same for the `credits` topic in `isetup.xml`, and later in `whatsnew.htm`.

Would you be able to replace the header like this?

martijnlaan:
I've finished adding a zstd branch to the iscompress repo. See https://github.com/jrsoftware/iscompress/compare/main...zstd for the work done. Most importantly:

- Downloaded Zstd 1.5.7 from [GitHub](https://github.com/facebook/zstd/archive/refs/tags/v1.5.7.zip), were same as your files, except that I removed the dictBuilder files.
- Initially copied your vcxproj files, but then instead redid isunzstd based on iszlib, and iszstd based on islzma.
- isunzstd is now CRT free, and it like zlib and bzip2 it uses a helper for this: [iszstd.c](https://github.com/jrsoftware/iscompress/blob/8ef9cea5e8cb7fe7b1e71d44c53fba7136c46325/zstd/isunzstd.c).
- Your isunzstd.vcxproj referenced pool.c and threading.c and defined ZSTD_MULTITHREAD=1, but zstd decompression isn't multithreaded as far as I can tell. So these files were unused, and now no longer referenced.
- Added iszstd-Arm64EC.dll.
- No .rc and .props files.

For the PR this means:
- Components\Zstd should be removed
- Projects\Src\Compression.Zstd should be removed
- The compiler should use iszstd-Arm64EC.dll same way it uses islzma-Arm64EC.dll.

amyspark:
Hi!

> isunzstd is now CRT free, and it like zlib and bzip2 it uses a helper for this: [iszstd.c](https://github.com/jrsoftware/iscompress/blob/8ef9cea5e8cb7fe7b1e71d44c53fba7136c46325/zstd/isunzstd.c).

It doesn't pass linking on Debug, I think you need to enable `/Oi` for all configurations (sorry for the logs in Spanish, I built this from within Visual Studio):

```
2>zstd_decompress.obj : error LNK2001: símbolo externo _byteswap_ulong sin resolver
2>zstd_decompress_block.obj : error LNK2001: símbolo externo _byteswap_ulong sin resolver
2>zstd_ddict.obj : error LNK2001: símbolo externo _byteswap_ulong sin resolver
2>entropy_common.obj : error LNK2001: símbolo externo _byteswap_ulong sin resolver
2>xxhash.obj : error LNK2001: símbolo externo _byteswap_ulong sin resolver
2>fse_decompress.obj : error LNK2001: símbolo externo _byteswap_ulong sin resolver
2>huf_decompress.obj : error LNK2001: símbolo externo _byteswap_ulong sin resolver
2>zstd_decompress_block.obj : error LNK2001: símbolo externo _byteswap_uint64 sin resolver
2>xxhash.obj : error LNK2001: símbolo externo _byteswap_uint64 sin resolver
2>fse_decompress.obj : error LNK2001: símbolo externo _byteswap_uint64 sin resolver
2>huf_decompress.obj : error LNK2001: símbolo externo _byteswap_uint64 sin resolver
2>zstd_decompress.obj : error LNK2001: símbolo externo _byteswap_uint64 sin resolver
2>xxhash.obj : error LNK2019: símbolo externo _rotl sin resolver al que se hace referencia en la función ZSTD_XXH32_digest
2>xxhash.obj : error LNK2019: símbolo externo _rotl64 sin resolver al que se hace referencia en la función ZSTD_XXH64_digest
```

Also, would it be possible to fix the output folder for that particular type so that it also gets output within the project folder?

Apart from that, I'm amazed at the reduction in size -- will check out now how it works!

amyspark:
@martijnlaan The Arm64EC output of isunzstd also needs fixing, it's going to the x64 folder with an x64 name. Is that correct?

martijnlaan:
There should be only iszstd-Arm64EC.dll, and not isunzstd-Arm64EC.dll. This is like islzma.

In other words, the 64-bit compiler should prefer to use the Arm64EC binary on Window 11 ARM and the x64 binary otherwise, but 64-bit Setup should always contain and use x64 binary only.

So don't try to compile Arm64EC isunzstd, even if it's partially present. Doesn't work and isn't supposed to.

Not sure what you mean with the question about output folders. Release build output should go to iscompress folders, debug output to issrc folders. It already is like that isnt it?

amyspark:
> Not sure what you mean with the question about output folders. Release build output should go to iscompress folders, debug output to issrc folders. It already is like that isnt it?

Ah, ah, I understood both went to iscompress. So that's OK.

Testing the initialization code as we speak, I should be able to test things in the next few hours.

martijnlaan:
> It doesn't pass linking on Debug, I think you need to enable `/Oi` for all configurations (sorry for the logs in Spanish, I built this from within Visual Studio):

I have fixed this. Needs /Oi indeed, was enabled by default in Release.

Will you also reply to the Compression.Zstd.pas question? https://github.com/jrsoftware/issrc/pull/622#issuecomment-5069488300

amyspark:
> Will you also reply to the Compression.Zstd.pas question? https://github.com/jrsoftware/issrc/pull/622#issuecomment-5069488300

No problem re: change of header + license, I'll push that and the removal of the Zstd code in the next commit. Shall I squash the branch, to keep it tidy?

Also re: isunzstd, I think it's leaving performance on the table, noticed something odd when testing the installer. Will keep you posted.

martijnlaan:
> No problem re: change of header + license, I'll push that and the removal of the Zstd code in the next commit. Shall I squash the branch, to keep it tidy?

That's great, thanks in advance. Squashing would be nice yet, and/or making a new PR: I have added a zstd branch on this repo as well, and the PR should target that. Currently it has no extra commits, but I will commit the signed DLLs there later.

On iscompress I have also added a [test.bat](https://github.com/jrsoftware/iscompress/blob/zstd/test/test.bat) and [roundtrip.c](https://github.com/jrsoftware/iscompress/blob/zstd/test/roundtrip.c) which tests the DLLs.

amyspark:
Rebase and squash done, all that's needed now is to address the bits in "TODO / Questions / Need for improvement" above.

(Basically how to communicate the thread count and UseSolidCompression from SetupCompiler to the Compressor.)

amyspark:
> Also re: isunzstd, I think it's leaving performance on the table, noticed something odd when testing the installer. Will keep you posted.

Would it be possible to enable LTCG on the iszstd project, maybe on isunzstd too? I got 4MB/s compression speedup with it enabled. (iszstd.c needs to have link-time code generation disabled as it's a shim)

martijnlaan:
I'm not sure what you mean. This is about `<WholeProgramOptimization>` right?

> Would it be possible to enable LTCG on the iszstd project

iszstd already has that?

afbeelding

afbeelding


> maybe on isunzstd too

isunzstd has it enabled on the solution level, but disabled on the CLCompile level. This is related to being CRT free.

> (iszstd.c needs to have link-time code generation disabled as it's a shim)

iszstd.c doesn't exist

martijnlaan:
> Rebase and squash done, all that's needed now is to address the bits in "TODO / Questions / Need for improvement" above.
> 
> (Basically how to communicate the thread count and UseSolidCompression from SetupCompiler to the Compressor.)

I'm not sure if this answers the question but `TLZMACompressorProps` shows various settings being communicated like thread counts. 

Solid compression is not a setting like that. Disabling solid compression causes the compiler to call FinalizeChunk early, regardless of the compression algorithm.

amyspark:
> I'm not sure if this answers the question but TLZMACompressorProps shows various settings being communicated like thread counts.

Can you have a look at the most recent commit, see if that's OK for passing Zstd properties?

Projects/Src/Compiler.SetupCompiler.pas
```diff
@@ -641,41 +673,38 @@ procedure TSetupCompiler.InitBzipDLL;
...
-    GetNativeSystemInfo(SysInfo);
-    if SysInfo.wProcessorArchitecture = PROCESSOR_ARCHITECTURE_ARM64 then
-      Exit(True);
-    {$ENDIF}
-
-    Result := False;
-  end;
-  
+{$IFDEF WIN64}
+function TSetupCompiler.IsArm64: Boolean;
```
martijnlaan: Should be moved to `Compiler.HelperFunc` instead

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,381 @@
...
+    raise ECompressInternalError.CreateFmt(SZlibInternalError, [Code]);
+  Result := True;
+end;
+
+{ TZstdCompressor }
+
+constructor TZstdCompressor.Create(AWriteProc: TCompressorWriteProc;
+  AProgressProc: TCompressorProgressProc; CompressionLevel: Integer;
+  ACompressorProps: TCompressorProps);
+var
```
martijnlaan, line 168: inline const wherever possible is preferred for new code, otherwise inline var. Only expections are arrays which dont work as inline in Delphi 10.4. Using inline const/var reduces risk of mistakes.

Projects/Src/IDE.ScintStylerInnoSetup.pas
```diff
@@ -2194,6 +2197,9 @@ function GetCompressionValues: TArray<TScintRawString>;
     for var Level in  LZMALevels do
       SetResult(I, TScintRawString(Algo + '/' + Level));
   end;
+  // SetResult(I, TScintRawString(ZstdAlgo));
```
martijnlaan: ?

Projects/Src/IDE.ScintStylerInnoSetup.pas
```diff
@@ -2268,10 +2274,11 @@ initialization
     SSDV(ssArchiveExtraction, ['auto', 'basic', 'enhanced/nopassword', 'enhanced', 'full']),
     SSDV(ssCloseApplications, ['force', SYes, SNo]),
     SSDV(ssCompression, GetCompressionValues),
-    SSDV(ssDisablePrecompiledFileVerifications, ['setup', 'setupcustomstyle', 'setupldr', 'is7z', 'isbunzip', 'isunzlib', 'islzma']),
+    SSDV(ssDisablePrecompiledFileVerifications, ['setup', 'setupcustomstyle', 'setupldr', 'is7z', 'isbunzip', 'isunzlib', 'islzma', 'isunzstd']),
     SSDV(ssEncryption, ['full', SYes, SNo]),
     SSDV(ssInternalCompressLevel, ['none'] + LZMALevels), { We don't list 0 }
     SSDV(ssLanguageDetectionMethod, ['uilanguage', 'locale', 'none']),
+
```
martijnlaan: Extra blank line added

Projects/Src/Shared.Struct.pas
```diff
@@ -81,7 +81,7 @@ interface
       64-bit builds, even without specifying packed. But to be sure we specify
       it anyway. }
   TSetupLanguageDetectionMethod = (ldUILanguage, ldLocale, ldNone);
-  TSetupCompressMethod = (cmStored, cmZip, cmBzip, cmLZMA, cmLZMA2);
+  TSetupCompressMethod = (cmStored, cmZip, cmBzip, cmLZMA, cmLZMA2, cmZstd);
```
martijnlaan, line 84: This means SetupID should be bumped, recommend using `'Inno Setup Setup Data (7.1.0)'` for now.

build-precomp.bat
```diff
@@ -121,4 +121,19 @@ copy %R%\x64\Release\islzma64.exe Files
 if errorlevel 1 exit /b 1
 copy %R%\Arm64EC\Release\islzma-ARM64EC.dll Files
 if errorlevel 1 exit /b 1
-exit /b
\ No newline at end of file
+exit /b
```
martijnlaan: This should be reverted. Dont think it needs another change now that iscompress build.bat builds the zstd files

martijnlaan:
> Can you have a look at the most recent commit, see if that's OK for passing Zstd properties?

It looks a bit odd, so probably not. I would have to better understand what the zsdt requires. What is all of the input that it accepts and what are the conditions?

As I understand it now, it accepts a thread count, which must be 1 if solid compression is used? If that is correct, then it should use CompressionThreads for that, and the compiler should enforce it begin 1 if solid compression is used.

Doesn't it have other input.

Something else we will need to know for later is the memory requirements, both for compression and decompression, and how this relates (or doesn't relate) to the compress level, the thread count, and any other settings.

martijnlaan:
BTW I updated the PR to point to the zstd branch instead of main.

martijnlaan:
> > Can you have a look at the most recent commit, see if that's OK for passing Zstd properties?
> 
> It looks a bit odd, so probably not. I would have to better understand what the zsdt requires. What is all of the input that it accepts and what are the conditions?

Looked at again and also how the compiler works, and I recommend you don't worry about this bit for now, I will do refactoring later. The rest of my comment still applies, so:

What is all of the input that it accepts and what are the conditions?

As I understand it now, it accepts a thread count, which must be 1 if solid compression is used? If that is correct, then it should use CompressionThreads for that, and the compiler should enforce it begin 1 if solid compression is used.

Doesn't it have other input.

Something else we will need to know for later is the memory requirements, both for compression and decompression, and how this relates (or doesn't relate) to the compress level, the thread count, and any other settings.

martijnlaan:
See f64f3035 and 137328a1

Projects/Src/IDE.ScintStylerInnoSetup.pas
amyspark: Ah, that was me tracking down a reported AnsiString leak. I'll fix.

amyspark:
> Something else we will need to know for later is the memory requirements, both for compression and decompression, and how this relates (or doesn't relate) to the compress level, the thread count, and any other settings.

I only tested variations in compression level, 8 threads, `SolidCompression=yes`, and my 5GB SDK. Everything else was left to libzstd's internally calculated defaults.

Level | Decompression (single threaded) | Compression
-- | -- | --
3 | 2,5 MB | 97 MB 
7 | 2,5 MB | 115 MB 
11 | 4,5 MB | 430 MB 
15 | 4,5 MB | 852 MB 
18 | 8,5 MB | 1067 MB 
21 | 64,5 MB | 7699 MB 
22 | 128,5 MB | 16410 MB 

3 is libzstd's default; 7 and 12 are some intermediates (with strategy changes); 18 is my default as in my tarball tests it's time-equivalent to XZ, and 21 and 22 are the extreme.

In all cases the install time was around 40s, compression time varied from ~1m10s (level 3) all the way to ~12m (level 22). My default of 18 takes ~3m50s.

Projects/Src/Compiler.SetupCompiler.pas
```diff
@@ -2693,6 +2688,7 @@ procedure TSetupCompiler.EnumSetupProc(const Line: PChar; const Ext: Integer);
         4: Include(Result, pfIsbunzip);
         5: Include(Result, pfIsunzlib);
         6: Include(Result, pfIslzma);
+        7: Include(Result, pfIszstd);
```
jordanrussell: Should that be `pfIsunzstd`?

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+  TZstdCompressor = class(TCustomCompressor)
+  private
+    FCompressionLevel: Integer;
+    FNumThreads: Integer;
+    FInitialized: Boolean;
+    FStrm: Pointer;
+    FOut: TZSTD_outBuffer;
+    { Zstd blocks are 128KB + overhead; give it 256KB to minimize roundtrip
+      times. See Destroy as to why this is important }
+    FBuffer: array[0..$40000] of Byte;
```
jordanrussell: Should that be `$3FFFF`?

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+  ZSTD_e_continue = 0;
+  ZSTD_e_flush    = 1;
+  ZSTD_e_end      = 2;
+
+  ZSTD_c_nbWorkers        = 400;
+
+  ZSTD_reset_session_only = 1;
+
+var
+  ZSTD_createCStream: function (): Pointer; stdcall;
```
jordanrussell: Please omit `()`

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+  end;
+  if FNumThreads = 0 then begin
+    { Let's make Zstd use automatically all physical processors }
+    GetActiveProcessorCountFunc := GetProcAddress(GetModuleHandle(kernel32),
+      'GetActiveProcessorCount');
+    if Assigned(GetActiveProcessorCountFunc) then begin
+      const ActiveProcessorCount = GetActiveProcessorCountFunc(65535);
+      if ActiveProcessorCount > 1 then
+        FNumThreads := ActiveProcessorCount div 2;
+    end;
```
jordanrussell, line 187: Is the `div 2` meant to exclude logical processors (SMT)? What if the CPU doesn't support SMT or has it disabled?

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+end;
+
+destructor TZstdCompressor.Destroy;
+begin
+  { Unlike other destructors e.g. Zlib, this library is backed by a runtime
+    thread pool. For this reason the buffer needs to be as small as
+    possible (1 block == 64KB), so as to trigger the flush ASAP, then every
+    write afterwards is promptly discarded from memory.
+    This helps a lot with returning the control to the event loop, minimizing
+    the time the UI spends unresponsive. }
```
jordanrussell, line 199: Calling `ProgressProc` here is unsafe as it may (legally) raise exceptions. On Delphi, destructors should never raise exceptions because it's likely to cause other destruction code in the calling function to be skipped, resulting in memory leaks or undefined behavior.

It's unclear to me, though, why that loop is necessary. Does zstd really not have a "cancel compression now" function?

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+  InitCompress;
+  var FIn: TZSTD_inBuffer;
+  FIn.src := @Buffer;
+  FIn.size := Count;
+  FIn.pos := 0;
+  var OldCount := FProgress.consumed;
+  while FIn.pos < Count do begin
+    Check(ZSTD_compressStream2(FStrm, FOut, FIn, ZSTD_e_continue));
+    if FOut.pos > 0 then
+      FlushBuffer;
```
jordanrussell, line 276: Shouldn't the buffer only be flushed when full (like `TZCompressor.DoCompress`)?

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+  if FOut.pos > 0 then begin
+    WriteProc(FBuffer, FOut.pos);
+    FOut.pos := 0;
+  end;
+end;
+
+procedure TZstdCompressor.DoCompress(const Buffer; Count: Cardinal);
+begin
+  InitCompress;
+  var FIn: TZSTD_inBuffer;
```
jordanrussell: Only member variables should have the `F` prefix. Local variables should have no prefix, or if having no prefix would conflict with another variable or reserved word (e.g., `in` is a reserved word), then an `L` prefix may be used.

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+end;
+
+{ TZstdDecompressor }
+
+constructor TZstdDecompressor.Create(AReadProc: TDecompressorReadProc);
+begin
+  inherited Create(AReadProc);
+  FStrm := ZSTD_createDStream();
+  Check(ZSTD_initDStream(FStrm));
+  FillChar(FIn, SizeOf(FIn), 0);
```
jordanrussell: Just to avoid duplication, looks like the above two lines could be replaced with `Reset;`

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+  FOut.size := Count;
+  FOut.pos := 0;
+  while FOut.pos < Count do begin
+    if FReachedEnd then  { unexpected EOF }
+      raise ECompressDataError.Create(SZlibDataError);
+    if FIn.pos = FIn.size then begin
+      FIn.src := @FBuffer;
+      FIn.size := ReadProc(FBuffer, SizeOf(FBuffer));
+      FIn.pos := 0;
+    end;
```
jordanrussell, line 352: If `ReadProc` were to return zero when zstd is still expecting more input, would this loop endlessly?

The LZMA and bzlib decompressor classes needed special handling for that situation (but raising an exception any time `ReadProc` returns zero as the bzlib class does may be wrong).

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+
+type
+  TZSTD_inBuffer = record
+    src: Pointer;        { start of input buffer }
+    size: UInt64;        { size of input buffer }
+    pos: UInt64;         { position where reading stopped. Will be updated. Necessarily 0 <= pos <= size }
+  end;
+
+  TZSTD_outBuffer = record
+    src: Pointer;        { start of output buffer }
```
martijnlaan: Should be dst

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+uses
+  Windows, SysUtils, Compression.Base, Compression.CompressorProps;
+
+function ZstdInitCompressFunctions(Module: HMODULE): Boolean;
+function ZstdInitDecompressFunctions(Module: HMODULE): Boolean;
+
+type
+  TZSTD_inBuffer = record
+    src: Pointer;        { start of input buffer }
+    size: UInt64;        { size of input buffer }
```
martijnlaan: This field is size_t in the C code, which means you should use NativeUInt, to make sure it's 4 bytes on 32-bit and 8 bytes on 64-bit, like size_t. Also applies to pos, TZSTD_outBuffer, returns types of the functions like ZSTD_initCStream, and ZSTD_isError res parameter.

Does not apply to TZSTD_frameProgression because for those the C code uses long long.

Projects/Src/Compiler.ExeUpdateFunc.pas
```diff
@@ -869,9 +869,7 @@ procedure UpdateIconsAndStyle(const FileName: String; const Uisf: TUpdateIconsAn
     end else if SameText(StyleFileName, 'builtin:slate') then
       StyleName := 'SLATECLASSICO'
     else if SameText(StyleFileName, 'builtin:stellar') then
-      StyleName := 'STELLAR'
```
martijnlaan, line 872: Should be removed from the PR. I will try and think of a way to better support compilation with the Community Edition.

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+begin
+  ZSTD_createCStream := GetProcAddress(Module, 'ZSTD_createCStream');
+  ZSTD_CCtx_setParameter := GetProcAddress(Module, 'ZSTD_CCtx_setParameter');
+  ZSTD_initCStream := GetProcAddress(Module, 'ZSTD_initCStream');
+  ZSTD_compressStream2 := GetProcAddress(Module, 'ZSTD_compressStream2');
+  ZSTD_freeCStream := GetProcAddress(Module, 'ZSTD_freeCStream');
+  ZSTD_isError := GetProcAddress(Module, 'ZSTD_isError');
+  ZSTD_CCtx_reset := GetProcAddress(Module, 'ZSTD_CCtx_reset');
+  ZSTD_getFrameProgression := GetProcAddress(Module, 'ZSTD_getFrameProgression');
+  Result := Assigned(ZSTD_createCStream) and Assigned(ZSTD_CCtx_setParameter) and Assigned(ZSTD_compressStream2) and
```
martijnlaan: ZSTD_initCStream check is missing

Projects/Src/Compiler.Messages.pas
```diff
@@ -114,6 +114,7 @@ interface
   SCompilerArchitectureIdentifierDeprecatedWarning = 'Architecture identifier "%s" is deprecated. ' +
     'Substituting "%s", but note that "%s" is preferred in most cases. See the "Architecture Identifiers" topic in help file for more information.';
   SCompilerRegTypeLibArchitectureMismatch = '%d-bit Setup cannot register type libraries in the %d-bit registry view. If you intend for the type library to be registered in the %0:d-bit registry view, add the "%0:dbit" flag';
+  SCompilerZstdTooManyThreads = 'NumThreads should be set to a fixed number when using Zstd compression with SolidCompression=no. More than one thread may lead to poor compression speeds with small files.';
```
martijnlaan: There's no NumThreads directive.

For now I suggest we use the CompressionThreads directive, seems the most natural fit.

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+  ACompressorProps: TCompressorProps);
+var
+  GetActiveProcessorCountFunc: function(GroupNumber: WORD): WORD; stdcall;
+begin
+  inherited;
+  FCompressionLevel := CompressionLevel;
+  FNumThreads := 0;
+  if ACompressorProps is TLZMACompressorProps then begin
+    const Props = (ACompressorProps as TLZMACompressorProps);
+    if Props.NumBlockThreads <> 0 then
```
martijnlaan, line 176: Should check NumThreads instead. The SCompilerZstdTooManyThreads already correctly checks NumThreads instead of NumBlockThreads.

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+    This helps a lot with returning the control to the event loop, minimizing
+    the time the UI spends unresponsive. }
+  var FUnusedIn: TZSTD_inBuffer;
+  FillChar(FUnusedIn, SizeOf(FUnusedIn), 0);
+  FOut.pos := 0;
+  var OldCount := FProgress.consumed;
+  var Code: UInt64 := 1;
+  if FProgress.consumed <> FProgress.ingested then begin
+    While Code <> 0 do begin
+      Code := ZSTD_compressStream2(FStrm, FOut, FUnusedIn, ZSTD_e_flush);
```
martijnlaan, line 207: Discards the bytes it flushes. The loop fills FBuffer, then does FOut.pos := 0 with no FlushBuffer/WriteProc call. Whatever it produces is dropped.

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+end;
+
+procedure TZstdCompressor.InitCompress;
+begin
+  { Decoupling initialization from compression context creation allows
+    reusing the context for further compression operations. Also, in
+    multithreaded mode, it's pretty easy to OOM Delphi by using Zstd together
+    with SolidCompression=no, as Zstd allocates a thread pool per context }
+  if FStrm = nil then begin
+    FStrm := ZSTD_createCStream();
```
martijnlaan, line 228: nil check missing

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+  end;
+  EndCompress;
+end;
+
+{ TZstdDecompressor }
+
+constructor TZstdDecompressor.Create(AReadProc: TDecompressorReadProc);
+begin
+  inherited Create(AReadProc);
+  FStrm := ZSTD_createDStream();
```
martijnlaan, line 316: nil check missing

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+    constructor Create(AReadProc: TDecompressorReadProc); override;
+    destructor Destroy; override;
+    procedure DecompressInto(var Buffer; Count: Cardinal); override;
+    procedure Reset; override;
+  end;
+
+implementation
+
+const
+  SZlibDataError = 'zstd: Compressed data is corrupted';
```
martijnlaan: Zlib -> Zstd

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
...
+begin
+  var FIn: TZSTD_inBuffer;
+  InitCompress;
+  var FReachedEnd := False;
+  FillChar(FIn, SizeOf(FIn), 0);
+  var OldCount := FProgress.consumed;
+  while not FReachedEnd do begin
+    const Code = ZSTD_compressStream2(FStrm, FOut, FIn, ZSTD_e_end);
+    if ZSTD_IsError(Code) <> 0 then
+      raise ECompressDataError.CreateFmt(SZlibInternalError, [Code]);
```
martijnlaan: ECompressDataError is for decompression, not for compression
Should be ECompressInternalError. Check has is right already, and I think it should just call Check here

Projects/Bin/synch-isfiles.bat
```diff
@@ -14,6 +14,8 @@ copy ..\..\Files\isscint*.dll
 copy ..\..\Files\isscint*.dll.issig
 copy ..\..\Files\islzma*.dll
 copy ..\..\Files\islzma*.dll.issig
+copy ..\..\Files\isunzstd*.dll
```
martijnlaan, line 17: .issig files copy is missing

Projects/Src/IDE.ScintStylerInnoSetup.pas
martijnlaan: Still there?

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,362 @@
+unit Compression.Zstd;
+
+{
+  Inno Setup
+  Copyright (C) 1997-2025 Jordan Russell
```
martijnlaan: 2025->2026

Projects/ISCmplr.dproj
```diff
@@ -167,6 +168,7 @@
         <DCCReference Include="..\Components\UniPs\Source\uPSC_stdctrls.pas"/>
         <DCCReference Include="..\Components\UniPs\Source\uPSCompiler.pas"/>
         <DCCReference Include="..\Components\UniPs\Source\uPSUtils.pas"/>
+        <DCCReference Include="Src\Compression.CompressorProps.pas"/>
```
martijnlaan: Not in correct position / mismatch with the dpr

martijnlaan:
For CompressionProps I recommend the following for now:
- Revert addition of `Compression.CompressorProps.pas`
- Revert removal of `TLZMACompressorProps`
- Add new `TThreadedCompressorProps` to` Compression.Base.pas`, as below
- Change `TLZMACompressorProps`: `TLZMACompressorProps = class(TThreadedCompressorProps)`
- Also remove `NumThreads` from it
- Now `Compression.Zstd.pas` can use:

```
if ACompressorProps is TThreadedCompressorProps then
  const Props = (ACompressorProps as TThreadedCompressorProps);
```

New TThreadedCompressorProps:
```
TThreadedCompressorProps = class(TCompressorProps)
public
  NumThreads: Integer;
end;
```

Ultimately the compiler would need a larger refactor for proper flexibility, but this is at least a step in the right direction, avoiding the refactor for now.

The ssCompressionThreads handling of auto and actual values will need fixing as well, but I should probably do that myself later. That would also remove SCompilerZstdTooManyThreads, because the compiler itself would then make things right.

jordanrussell:
I don't understand the issue with having multiple threads enabled while solid compression is disabled. LZMA also supports using extra threads (`LZMANumBlockThreads`), and the compression is also done asynchronously (`DoCompress` dumps the input bytes into a buffer and returns). With `LZMANumBlockThreads=2` and `UseSolidCompression=no`, there likely is a minor slowdown if all files are tiny -- the block threads are created/destroyed for each file. But if the files are large, and the user wants fast random access to files, then that combination of settings may make perfect sense. The compiler doesn't issue warnings about suboptimal settings because it doesn't know what is optimal for the user's file set.

For consistency with LZMA, though, it would probably be best to stick with one thread by default -- assuming zstd's extra threads are equivalent to LZMA's "block threads" (I haven't looked into it).

Projects/Src/Compression.Zstd.pas
amyspark: > If ReadProc were to return zero when zstd is still expecting more input, would this loop endlessly?

Yes, once the decompression buffer is exhausted (but the frame is not finished) it will return > 0 until more data is supplied.

Projects/Src/Compression.Zstd.pas
amyspark: > Is the div 2 meant to exclude logical processors (SMT)? What if the CPU doesn't support SMT or has it disabled?

Yes. As explained in the PR's description, I had no way to communicate an actual number of threads to libzstd. Even now, when using `CompressionThreads`, it only allows `auto` (which I assume means "use as many as possible") or `1` (compress in a separate thread, then wait on `DoFinish`). So I set it to the (supposed) number of cores so as to have the maximum possible, but not risking CPU starvation or BSoDs.

(My R7 2700x tends to lock up whenever facing 100% CPU core usage with AVX2, but that's not relevant for iszstd as it's not built with `/arch:AVX2`.)

Projects/Src/Compression.Zstd.pas
amyspark: > It's unclear to me, though, why that loop is necessary. Does zstd really not have a "cancel compression now" function?

`ZSTD_freeCStream`, or alternatively `ZSTD_compressStream2` with endDirective = `ZSTD_e_flush` as used here. However, under the hood `ZSTD_freeCStream`  conducts the same steps (consume remaining input data, then discard compressed buffers, close thread pool and free the memory) but *without* allowing any sort of callback to not block the event loop. You can easily lock ISIDE up for ages if you're using high compression levels, say >= 19, together with `SolidCompression=yes`. That's why I do it manually here, to keep the event loop responsive.

Projects/Src/Compression.Zstd.pas
amyspark: Correct, see comment earlier as to why this is needed. (It's still done internally in case of outright context free)

Projects/Src/Compression.Zstd.pas
amyspark: Yeah, this can be optimized; it was a leftover from when I didn't know about `ZSTD_getFrameProgression`.

amyspark:
> For consistency with LZMA, though, it would probably be best to stick with one thread by default -- assuming zstd's extra threads are equivalent to LZMA's "block threads" (I haven't looked into it).

Zstd's threads manage up to `overlapSize` bytes, which is itself up to a `windowSize` (at least 8MB apiece IIUC).

re isunzstd @martijnlaan : your UCRT-less refactor almost doubles the decompression time I'd achieved, but I found there's a workaround for that. Would it be possible to piggy back on kernel32's `Rtl*` functions? That way I can get the current 40-43s install time down to 24-27s.

martijnlaan:
> re isunzstd @martijnlaan : your UCRT-less refactor almost doubles the decompression time I'd achieved, but I found there's a workaround for that. Would it be possible to piggy back on kernel32's `Rtl*` functions? That way I can get the current 40-43s install time down to 24-27s.

Yeah, I would like to see which functions that are. You could make a PR on iscompress's zstd branch? Keep in mind that it needs to support Windows 7 and later (aka Windows Server 2008 R2 and later).

amyspark:
> The ssCompressionThreads handling of auto and actual values will need fixing as well, but I should probably do that myself later. That would also remove SCompilerZstdTooManyThreads, because the compiler itself would then make things right.

Feel free to address this @martijnlaan; for the time being I'll leave the processor count query API logic as a placeholder. I think it'll be best to move `NumBlockThreads` to the threaded props instead, or to uncap `CompressionThreads` depending on the active compressor (unsure if that is possible).

martijnlaan:
> or to uncap `CompressionThreads` depending on the active compressor (unsure if that is possible).

It should be. `ssCompressionThreads` handling should just store the input into a regular variable, and then somewhere after` { Verify settings set in [Setup] section }` it would set actual `CompressProps.NumThreads` based on that + `SetupHeader.CompressMethod`.

But I would first like to understand how Jordan thinks this all should work (not the code, but what directives and values).

Projects/Src/Compression.Zstd.pas
jordanrussell: I see... How long is "ages"? 5 seconds? 10 seconds? Longer?

jordanrussell:
`CompressionThreads` (and the LZMA `numThreads` variable it sets) really isn't the "number of threads" used, but rather effectively a Boolean value specifying whether the LZMA MT match finder should be enabled, where `1=false` and `2=true`. Exposing it as a `yes`/`no` setting would have made more sense in retrospect, but I think I was planning for the possibility of LZMA accepting other `numThreads` values in the future (like 3).

It appears that zstd's worker threads are equivalent to LZMA's block threads, which are enabled via `LZMANumBlockThreads`. So, I think there ultimately should be a zstd-specific directive like `ZstdWorkerThreads` (and a separate `TZstdCompressorProps` class). But for the time being, `LZMANumBlockThreads` (`NumBlockThreads` in the code) could affect both algorithms.

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,366 @@
...
+      raise ECompressDataError.Create(SZstdDataError);
+    if FIn.pos = FIn.size then begin
+      FIn.src := @FBuffer;
+      FIn.size := ReadProc(FBuffer, SizeOf(FBuffer));
+      FIn.pos := 0;
+    end;
+    const Code = ZSTD_decompressStream(FStrm, FOut, FIn);
+    if ZSTD_isError(Code) <> 0 then begin
+      raise ECompressDataError.CreateFmt(SZstdInternalError, [Code]);
+    end else if Code = 0 then begin
```
jordanrussell: `ZSTD_decompressStream` docs say:
```
@return : 0 when a frame is completely decoded and fully flushed,
```
But a stream may be comprised of multiple frames, so `Code = 0` doesn't necessarily mean "reached end"?

Projects/Src/Compression.Zstd.pas
jordanrussell: Okay, so if I understand correctly, it needs a "was nothing consumed and nothing flushed?" check like `TLZMADecompressor` has:
```
  // before call:
  const OldInPos = FIn.pos;
  const OldOutPos = FOut.pos;
  // after call:
  if (FIn.pos = OldInPos) and (FOut.pos = OldOutPos) then
    raise ECompressDataError.Create(SZstdDataError);
```

martijnlaan:
I've updated the signed zstd binaries in 02ce78f8

Projects/ISCmplr.dproj
```diff
@@ -134,6 +134,8 @@
         <DCCReference Include="Src\Compression.Zlib.pas"/>
         <DCCReference Include="Src\Compression.bzlib.pas"/>
         <DCCReference Include="Src\Compression.LZMACompressor.pas"/>
+        <DCCReference Include="Src\Compression.Zstd.pas"/>
+        <DCCReference Include="Src\Compression.CompressorProps.pas"/>
```
martijnlaan: The projects still refer to remove CompressorProps, so they don't compile.

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,366 @@
...
+
+{ TZstdDecompressor }
+
+constructor TZstdDecompressor.Create(AReadProc: TDecompressorReadProc);
+begin
+  inherited Create(AReadProc);
+  FStrm := ZSTD_createDStream();
+  if FStrm = nil then
+    OutOfMemoryError;
+  Reset;
```
martijnlaan: If Reset raises then FStrm is not freed

Projects/Bin/synch-isfiles.bat
```diff
@@ -14,6 +14,9 @@ copy ..\..\Files\isscint*.dll
 copy ..\..\Files\isscint*.dll.issig
 copy ..\..\Files\islzma*.dll
 copy ..\..\Files\islzma*.dll.issig
+copy ..\..\Files\isunzstd*.dll
+copy ..\..\Files\isunzstd*.dll.issig
+copy ..\..\Files\iszstd*.dll
```
martijnlaan, line 19: .issig copy missing

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,366 @@
...
+  end;
+
+  TZstdDecompressor = class(TCustomDecompressor)
+  private
+    FInitialized: Boolean;
+    FStrm: Pointer;
+    FIn: TZSTD_inBuffer;
+    FReachedEnd: Boolean;
+    { Cache a few blocks more than the compressor }
+    FBuffer: array[0..$FFFFF] of Byte;
```
martijnlaan, line 74: FBuffer is actually same size as in compressor?

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,366 @@
...
+  end;
+
+  TZstdCompressor = class(TCustomCompressor)
+  private
+    FCompressionLevel: Integer;
+    FNumThreads: Integer;
+    FInitialized: Boolean;
+    FStrm: Pointer;
+    FOut: TZSTD_outBuffer;
+    { Zstd blocks are 128KB + overhead }
```
martijnlaan: Is it correct that this says 128 KB while next line is 1 MB?

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,366 @@
...
+    end;
+  end;
+  InitCompress;
+end;
+
+destructor TZstdCompressor.Destroy;
+begin
+  { Unlike other destructors e.g. Zlib, this library is backed by a runtime
+    thread pool. For this reason the buffer needs to be as small as
+    possible (1 block == 64KB), so as to trigger the flush ASAP, then every
```
martijnlaan: Line 51 says blocks are 128 KB.

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,366 @@
...
+  ZSTD_initCStream: function (zcs: Pointer; compressionLevel: Integer): NativeUInt; stdcall;
+  ZSTD_CCtx_setParameter: function(cctx: Pointer; param: Cardinal; value: Integer): NativeUInt; stdcall;
+  ZSTD_compressStream2: function(cctx: Pointer; var output: TZSTD_outBuffer; var input: TZSTD_inBuffer; endOp: Cardinal): NativeUInt; stdcall;
+  ZSTD_freeCStream: function(zcs: Pointer): NativeUInt; stdcall;
+
+  ZSTD_createDStream: function(): Pointer; stdcall;
+  ZSTD_initDStream: function(zds: Pointer): NativeUInt; stdcall;
+  ZSTD_decompressStream: function(zds: Pointer; var output: TZSTD_outBuffer; var input: TZSTD_inBuffer): NativeUInt; stdcall;
+  ZSTD_freeDStream: function(zds: Pointer): NativeUInt; stdcall;
+  ZSTD_sizeof_DStream: function(zds: Pointer): NativeUInt; stdcall;
```
martijnlaan: ZSTD_sizeof_DStream isnt used anywhere?

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,366 @@
...
+    This helps a lot with returning the control to the event loop, minimizing
+    the time the UI spends unresponsive. }
+  var FUnusedIn: TZSTD_inBuffer;
+  FillChar(FUnusedIn, SizeOf(FUnusedIn), 0);
+  FOut.pos := 0;
+  var OldCount := FProgress.consumed;
+  var Code: NativeUInt := 1;
+  if FProgress.consumed <> FProgress.ingested then begin
+    While Code <> 0 do begin
+      Code := ZSTD_compressStream2(FStrm, FOut, FUnusedIn, ZSTD_e_flush);
```
martijnlaan, line 207: ZSTD_isError isn't called?

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,366 @@
...
+    if FReachedEnd then  { unexpected EOF }
+      raise ECompressDataError.Create(SZstdDataError);
+    if FIn.pos = FIn.size then begin
+      FIn.src := @FBuffer;
+      FIn.size := ReadProc(FBuffer, SizeOf(FBuffer));
+      FIn.pos := 0;
+    end;
+    const Code = ZSTD_decompressStream(FStrm, FOut, FIn);
+    if ZSTD_isError(Code) <> 0 then begin
+      raise ECompressDataError.CreateFmt(SZstdInternalError, [Code]);
```
martijnlaan: This doesn't seem right. Corrupted data is not an internal error.

Projects/Src/Compression.Zstd.pas
martijnlaan: Not resolved? Still says src

Projects/Src/IDE.ScintStylerInnoSetup.pas
```diff
@@ -2283,4 +2289,4 @@ initialization
     SSDV(ssWizardImageAlphaFormat, ['none', 'defined', 'premultiplied']),
     SSDV(ssWizardStyle, ['classic', 'modern', 'light', 'dark', 'dynamic', 'excludelightbuttons', 'excludelightcontrols', 'includetitlebar', 'hidebevels', 'polar', 'slate', 'stellar', 'windows11', 'zircon'])];
 
-end.
+end.
```
martijnlaan: Newline added

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,366 @@
...
+    possible (1 block == 64KB), so as to trigger the flush ASAP, then every
+    write afterwards is promptly discarded from memory.
+    This helps a lot with returning the control to the event loop, minimizing
+    the time the UI spends unresponsive. }
+  var FUnusedIn: TZSTD_inBuffer;
+  FillChar(FUnusedIn, SizeOf(FUnusedIn), 0);
+  FOut.pos := 0;
+  var OldCount := FProgress.consumed;
+  var Code: NativeUInt := 1;
+  if FProgress.consumed <> FProgress.ingested then begin
```
martijnlaan: This makes assigning ProgressProc a requirement, else FProgress is never updated.

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,366 @@
...
+  TZstdCompressor = class(TCustomCompressor)
+  private
+    FCompressionLevel: Integer;
+    FNumThreads: Integer;
+    FInitialized: Boolean;
+    FStrm: Pointer;
+    FOut: TZSTD_outBuffer;
+    { Zstd blocks are 128KB + overhead }
+    FBuffer: array[0..$FFFFF] of Byte;
+    { Workaround for Zstd not resetting the frame progression until compress2
```
martijnlaan: Trailing whitespace

Projects/Bin/synch-isfiles.bat
```diff
@@ -35,4 +38,4 @@ echo - Synching files done
 
 if /I "%1"=="nopause" goto :eof
 
-pause
\ No newline at end of file
+pause
```
martijnlaan: Added newline

Projects/Src/Compression.Zstd.pas
```diff
@@ -0,0 +1,366 @@
...
+      const Actual = FProgress.consumed - OldCount;
+      ProgressProc(Actual);
+      OldCount := FProgress.consumed;
+    end;
+  end;
+end;
+
+procedure TZstdCompressor.DoFinish;
+begin
+  var FIn: TZSTD_inBuffer;
```
martijnlaan: Still F on local, also bit below for FReachedEnd

martijnlaan:
To test the Rtl functions, I have tested all binaries on Windows 7 SP1 and Wine 9.0 as well.

martijnlaan:
Now also with Wine 11.0.

Projects/Src/Compression.Zstd.pas
amyspark: Yeah, I'll remove this outright. It was meant for when I hadn't implemented frame progression reporting.

Projects/Src/Compression.Zstd.pas
amyspark: Yes. The purpose here is to feed as much data to libzstd as possible, as I can control the read size.

Projects/Src/Compression.Zstd.pas
amyspark: It was used for reporting the decompression context size in the MR; will remove, I shouldn't have committed it.

Projects/Src/Compression.Zstd.pas
amyspark: In my 5GB payload tests, at level 22 it's at least 5 minutes, as it will take less than 30s to read and buffer the complete folder into memory.

Projects/Src/Compression.Zstd.pas
amyspark: (do correct me if I'm mistaken, this is based on my understanding of how Inno's compression loop works)

Each roundtrip of DoCompress/DoFinish is one frame. So e.g. in `SolidCompression=yes` it's just the one frame throughout the stream, and with `SolidCompression=no` it's one frame per file.

If at any read it has more data than expected, then that's an error and gets checked here, because AIUI Inno doesn't add extra data to compressed streams and then supplies it blindly to decompressors.

Projects/Src/Compression.Zstd.pas
amyspark: Does the change I pushed look good? (My past experience is with Free Pascal, so I never used Delphi exceptions.)

amyspark:
Made an extra push because I forgot to commit some extra changes.

Projects/Src/Compression.Zstd.pas
martijnlaan: I meant: if the size is the same, then why does it say 'Cache a few blocks more than the compressor'

Projects/Src/Compression.Zstd.pas
martijnlaan: Yes it does, thanks. Only the `on E: Exception do` should be removed because it's not needed, and now suggests you will access to the exception itself, while you don't.

Projects/Src/Compression.Zstd.pas
amyspark: IIRC the compressor gets fed blocks of 64KB each, but as the frame is 128KB, I wanted to be able to have bigger reads where I could control them.

This line was removed in the last push as it made no relevance.

Projects/Src/Compression.Zstd.pas
amyspark: I guess there should be an (optional) procedure for letting the compressor know it should stop, and thus allow ProgressProc to be safely called. But how does Inno go from `FCompileWantAbort` to `CH.Free` at present? It's wrapped in a Try-Finally, so something must be triggering the Free (a swallowed exception?) but the Delphi IDE isn't telling me what's invoking the Finally block.

martijnlaan: FCompileWantAbort is in ISIDE. If its set to True next time ISCmplr calls ISIDE's callback (CompilerCallbackProc) with Code iscbNotifyIdle, then ISIDE will return iscrRequestAbort. Upon receiving that ISCmplr will call Abort, which raises an EAbort exception. This will then unwind everything, including running CH.Free. 

Most likely you have EAbort setup as an exception type to ignore, because that's the default I think.

Tools->Options->Debugger->Embarcadero Debuggers->Language Exceptions

amyspark: Would it be possible to add a `Stop` procedure in `TCompressionHandler` / `TCustomCompressor` so that it gets called within a nested `try..except` handler on `EAbort`?

I can file a separate PR if wanted.

martijnlaan: edit - nevermind, saw I repeated what [Amy already said](https://github.com/jrsoftware/issrc/pull/622#discussion_r3650583432) - hard to keep track of things using the GitHub UI only

Projects/Src/Compression.Zstd.pas
jordanrussell: Ah yes, it is indeed correct. I *thought* I had read somewhere that multi-threaded mode produced multiple frames, but no, no true.

Projects/Src/Compression.Zstd.pas
jordanrussell: Even if the UI is kept responsive while compression winds down, the IDE will still remain in a "Compiling" state for 5 minutes, during which time the user cannot edit, start another compile, or close the IDE. Right?

If there really is no way to tell zstd to "stop immediately" (I haven't tested myself), then it sounds like the wait could only be eliminated by moving the compression out-of-process where the compiler can kill it via `TerminateProcess`. But that's certainly out of scope for this PR. And I would hate to put a lot of effort into implementing that, only to find later that zstd fixes it on their end.

For now, until a better solution is found, how about just blocking access to the compression levels that cause the long wait (>=19?)?

martijnlaan: I think limiting the 'job size' would be better, at least by default. A limit of 32 MB (=level 19's auto job size) should make it stop quickly on 20-22 as well (but check the 'overlap size' stuff below). This way it would allow those to be used, though they would lose some ratio. The benefit is that it would also reduce memory use by a lot.

```pascal
if FNumThreads > 1 then
  if FCompressionLevel >= 20 then
    Check(ZSTD_CCtx_setParameter(FStrm, ZSTD_c_jobSize, 32 * 1024 * 1024));
```

Could be a directive, allowing job size to be overridden. Zstd clamps the value, so we wouldnt have to check for min/max ourselves.

Something like (pseudo code)
```pascal
if FNumThreads > 1 then begin
  if FCustomJobSize then
    Check(ZSTD_CCtx_setParameter(FStrm, ZSTD_c_jobSize, FCustomJobSize))
  else if FCompressionLevel >= 20 then
    Check(ZSTD_CCtx_setParameter(FStrm, ZSTD_c_jobSize, 32 * 1024 * 1024));
end;
```

It uses >= 20 above instead of >= 19 because 19 has 32 MB which should be small enough for quick stop. So the issue starts at 20 and not at 19.

There's one more thing though. There's also an 'overlap size' and the job size must be >= overlap size. This means zstd might ignore the specified job size:

At 20 the default overlap size is 32 MB, but at 21 it is 64 MB and at 22 it is 128 MB. So to fully enforce 32 MB job size at all levels (and at the cost or more ratio loss) it would also need to reduce the overlap size.

The way to set that is var ZSTD_c_overlapLog, but that's a 0-9 number. To get 32 MB at 21 it must be set to 8, to get 32 MB at 22 it must be set to 7. Could be another directive as well. It is also clamped by zstd.

**A simple option would be refusing >= 21 always, and forcing job size to be 32 MB at 20.**

martijnlaan:
@amyspark Will start working on this myself as well. If you have more changes to commit can you make those normal commits instead of squashing / force pushing? Thanks for all the work 👍

martijnlaan:
See:

35656f21 zstd: Reduce diff a bit and remove SCompilerZstdTooManyThreads.
4fb7d9a0 zstd: Update SetupID, wont be in 7.1.0 already.
37a5c042 zstd: Remove TZstdCompressor.Destroy's progress loop, to be replaced with limiting levels and/or job size. Avoid Check exceptions in destructors. Remove F on local variable. Remove 'empty' (). Fix W1072 warnings.
eeca910d zstd: Remove automatic use of all processors.
be1eb45f zstd: TZstdDecompressor Create/Destroy cleanup.
e81d9583 zstd: Fix two comments.
810d07b1 zstd: Bit of cleanup.

Now just need the decision on

1. Refuse level >= 19, OR:
2. Refuse level >= 21 and set job size to 32 MB at 20, OR:
3. ... ?

Also open

1. @jordanrussell asked "Should that be $3FFFF?" FBuffer, but both buffers are still array[0..$FFFFF] in both cases
2. ZstdNumWorkerThreads
3. Updating docs
4. ... ?

martijnlaan:
77fedaf7 zstd: Extra cleanup.

martijnlaan:
GitHub auto-closed this even though there are open points and I didnt mean to close it.

Also: 8dd9f6e0 zstd: Move IsArm64 back into SetupCompiler + deduplicate things.

martijnlaan:
01d4a051 zstd: Tweak formatting.
5cef43db zstd: Add ZstdNumWorkerThreads 1..64. Shares storage with LZMANumBlockThreads, which is a bit ugly but avoids a refactor.

Also looks like ZSTD_isError should be replaced with ZSTD_getErrorCode. This allows it to properly distinct between corruption and out of memory, like the others, especially important with decompression. Will make that so.

amyspark:
> Refuse level >= 21 and set job size to 32 MB at 20, OR:

Please set job size to 32MB. It made a *huge* dent on memory usage, making > 18 much more tractable for my CI. 

> @jordanrussell asked "Should that be $3FFFF?" FBuffer, but both buffers are still array[0..$FFFFF] in both cases

Each block is 128kb, so I went with 1MB to get a reasonable chunk data out of the (de)compression call before issuing WriteProc.

martijnlaan:
I've made it refuse 21 and 22, and forcing 32 MB job size at 19. I also made 19 the default. Also reduced the amount of levels suggested in autocomplete to the meaningful ones. 

Also renamed ZstdNumWorkerThreads to ZstdNumThreads, because it doesn't match 1-on-1 with zstd's worker threads:

At ZstdNum(Worker)Threads=1 it doesnt set ZSTD_c_nbWorkers so the default of 0 workers is used, zstd is in synchronous mode and there's 1 thread compressing (our thread). So ZstdNumWorkerThreads is a bad name, because ZstdNumWorkerThreads=1 gives 0 workers.

At ZstdNumThreads=2 it does set ZSTD_c_nbWorkers, zstd in in asynchronousmode, and there's 2 threads compressing: the 2 worker threads. So ZstdNumThreads is good name for both cases.

Also updated the Compression topic, and added a ZstdNumThreads topic.

And some minor cleanup.

Updating whatsnew and the credits section is to do but otherwise I think it should be done now.

martijnlaan:
Did a few tests with our own build.

`lzma/max` (default level) compressed slightly better than `zstd/19` (also default level) 
`zstd/19` was better than `lzma/max`
`lzma/ultra` was better than `zstd/19`

Also notice that the memory requirements of `lzma/max` and `zstd/19` are same, both for compression and decompression which I suppose is a sign that `19` is a good default level indeed.

amyspark:
> Also notice that the memory requirements of lzma/max and zstd/19 are same

@martijnlaan Mind checking with `lzma2/max` as well?

> So ZstdNumWorkerThreads is a bad name, because ZstdNumWorkerThreads=1 gives 0 workers.

IIRC that should be 1 worker in asynchronous mode, should make the UI more responsive.

martijnlaan:
> > Also notice that the memory requirements of lzma/max and zstd/19 are same
> 
> @martijnlaan Mind checking with `lzma2/max` as well?

Was just reporting what the table in the help says.

> > So ZstdNumWorkerThreads is a bad name, because ZstdNumWorkerThreads=1 gives 0 workers.
> 
> IIRC that should be 1 worker in asynchronous mode, should make the UI more responsive.

Don't think I agree. Using asynchronous mode with 1 worker would increase compression memory at 19 from 98 MB to 274 MB. Responsiveness is fine in synchronous mode for me.

martijnlaan:
Also, forgot to mention: using asynchronous mode hurts the compression ratio, even if using only 1 worker.  Normally this is negligble or even no change, but at levels 16-18 it can add 12.5% of the uncompressed size in the worst case. At lower levels more. If the uncompressed size is high but the compressed size is not, this hurts the ratio by lots and lots.

Tested it with a specially crafted 100 MB file: at level 18 in synchronous mode it compressed to 4 MB, and in asynchronous mode to almost 16MB.

BTW, your original code already used synchronous mode when the thread setting is 1, this is not new change.

martijnlaan:
I have merged this into the new `vnext` branch, which will be 7.2 (so not next release because that's 7.1).

whatsnew is now updated and also the contributors topic.

While working on this I did realize the way we now block levels 21 and 22 should not be needed in synchronous mode. Only in asynchronous mode (`ZstdNumThreads` > 1) the abort issue should appear. So I will change that later.

amyspark:
> I have merged this into the new vnext branch, which will be 7.2 (so not next release because that's 7.1).

Thanks so much!

PR #625, opened 2026-07-28
Update Italian.isl

bovirus:
@martijnlaan 

Please merge. Thanks.

### DIFF
--- a/Files/Languages/Italian.isl
+++ b/Files/Languages/Italian.isl
@@ -9,7 +9,7 @@
 ; Setup adds the periods automatically (appending a period would result in
 ; two periods being displayed).
 ;
-; Italian.isl - Last Update: 24.07.2026 by bovirus (bovirus@gmail.com)
+; Italian.isl - Last Update: 28.07.2026 by bovirus (bovirus@gmail.com)
 ;
 ; Translator name:   bovirus
 ; Translator e-mail: bovirus@gmail.com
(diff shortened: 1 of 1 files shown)


martijnlaan:
Thanks!

It looks like this update introduces deviations again.

1. SelectLanguageLabel should not end with a colon
2. ComponentsDiskSpaceGBLabel/ComponentsDiskSpaceMBLabel are not about free space, because they don't take already used space into account.
3. WelcomeLabel2, NoUninstallWarning, ReadyLabel1, PreparingDesc, and InstallingLabel all dropped the "on your computer"

I have fixed this in ae7ef7a3. Note that for WelcomeLabel2 and NoUninstallWarning I kept your new sentence structure and only restored "sul/nel computer".

But please for future updates: do **not** change meaning. The guidelines in Default.isl should be quite clear.

bovirus:
@martijnlaan

1. I checked in a previous version. I will check again.
2. Undesrtood if you already instaleld befroe don't need al space sued as free. OK.
3. "in the computer" for me is not necessary because don't add important info. In what other destination could install it if not in the computer?
Note: right sentence is "nel computer".

martijnlaan:
I don't want to be blunt, but as the translator your personal opinion about the English messages doesn't matter. If they say "on your computer" then the translation must include it. If they say "space" instead of "free space", then the translation must say "space". You must be able to accept this if you want to maintain the translation.

As a user your opinion about the English messages does matter, and you can post about them on the forum, or make suggestions, but mind that the bar to change anything about the English messages is very high, because of the time cost of handling this.

Your "nel computer" note is fine: "sul computer" versus "nel computer" doesn't change the meaning, so that choice is yours. Feel free to change WelcomeLabel2 to "nel computer" in your next update. It now says "sul computer" because that's what it said before you removed it.

PR #626, opened 2026-08-01
CONTRIBUTING.md: no squashing or force pushing in PRs

pzhlkj6612:
Hello! Maybe we need this notice for contributors?

I'm from https://github.com/jrsoftware/issrc/pull/622#issuecomment-5129660097 .


martijnlaan:
Thanks for the suggestion but early on in the PR the squashing was agreed on, because otherwise unwanted changes (addition of unneeded files) would have stayed in our history.

PR #627, opened 2026-08-06
Docs: Add note about ShouldSkipPage call frequency

M-Saadawy:
Adds a note clarifying that ShouldSkipPage may be called multiple times per page navigation, including for the current page when the user clicks Next. Recommends caching results in InitializeWizard to avoid expensive operations inside the function.

Verified via runtime probe of InnoSetup 7.0.2.


martijnlaan2:
Thanks. Added a more general note instead, based on the one there already was for check functions.

PR #629, opened 2026-08-22
Update Italian.isl

bovirus:
@martijnlaan 

Match english term (shortuct -> collegamento).

I follow the original english text.

Note: in this context could be more clear "collegamento programm" (=program shortuct).

I rember well that in a previous discussion that in Inno Setup 7 you will change "NameAndVersion=%1 version %2" with "NameAndVersion=%1 %2" to avoid on top left "version" between program name and version?

Thanks.


martijnlaan:
Thanks. `NameAndVersion` should not change. It is the message for script authors who do still want to include "version". Other uses of "version" have indeed been removed.

